Issue #31, opened 2021-05-18
Folder name in Mount point cannot have "."


AffiTheCreator:
### Reproduction Steps
During the application setup, under File System, when selecting the folder name to mount inside of the container if the name of the folder contains a "." saas boost won't let you apply the settings and will throw an error. (e.g /random.folder/ )

##### Step-by-step:
1. Configure your application
2. Tick **Provision a File system for the application**
3. Under "Mount point" try adding  `/random.folder`
4. Press the Submit button 

### What did you expect to happen?
It should have allowed me to create the folder. 

### What actually happened?
I got an error saying my folder name is invalid with no clue to what was happening. 

### Environment

  - **AWS Region  : eu-west-1**
  - **AWS SaaS Boost Version : 1.0.0?? not sure what version is on now ( my bottom corner says "Version v0 - vi )
  - **Workload OS  (Linux or Windows) : Linux **

This is :bug: Bug Report

hariohmprasath:
Hi @AffiTheCreator,
The following code will fix the issue and will fall inline with [EFS documentation](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html#enforce-root-directory-access-point) on the mount points

### ApplicationComponent.js

File path  - [https://github.com/awslabs/aws-saas-boost/blob/main/client/web/src/settings/ApplicationComponent.js](https://github.com/awslabs/aws-saas-boost/blob/main/client/web/src/settings/ApplicationComponent.js)

```javascript
Yup.string()
    .matches(/^(\/[a-zA-Z._-]+)*$/, "Invalid path. Ex: /mnt")
    .max(100, "The full path can't exceed 100 characters in length")
    .test("subdirectories", "The path can only include up to four subdirectories", val => (val.match(/\//g) || []).length <= 4)
    .required()
```

After making this change you can run `sh install.sh` from root directory and use option 3 (Update Web Application for existing AWS SaaS Boost deployment) to update existing application with this fix. I will do some validation and submit a pull request with these changes shortly.

~ Hari

PR #33, opened 2021-05-18
Fix for #31 #35 #39

hariohmprasath:
The following code will fix the issue and will fall inline with EFS documentation on the mount points:

[https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html#enforce-root-directory-access-point](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html#enforce-root-directory-access-point)

----


4patelr:
Tested fixes and closes #31, closes #35, and closes #39

PR #56, opened 2021-05-30
Change the code used to get accountId. - iam.getUser() to sts.getCallerIdentity()

skipio11:
Using sts.getCallerIdentity(), installer can work based on IAM Role when getting accountId.
So, in Cloud9, it can be easily applied without using the IAM user credentials directly.

----


4patelr:
This change has been incorporated in to the PR #65 so closing out.

Issue #82, opened 2021-06-24
Change log4j2 configuration to include log level in the installer


PoeppingT:
Currently the installer log4j configuration doesn't include the log level, which makes the installer log harder to parse at eye level. Ideally the log4j2 configuration in the installer would match the configuration for the other parts of SaaS Boost.

### Use Case

The installer is the first component a new user of SaaS Boost will interact with, so it should reflect the standards used in the rest of the SaaS Boost project.

### Proposed Solution

Change the installer log4j2 config in `installer/src/main/resources/log4j2.xml`

### Other

* [x] :wave: I may be able to implement this feature request
* [ ] :warning: This feature might incur a breaking change

---

This is a :rocket: Feature Request

PR #88, opened 2021-06-25
Standardize log4j2 configuration.

PoeppingT:
See https://logging.apache.org/log4j/2.x/manual/layouts.html#Patterns for detailed information on log4j2 Patterns.

Fixes https://github.com/awslabs/aws-saas-boost/issues/82

----


brtrvn:
Any reason not to make this log file match the pattern of the rest of SaaS Boost?
I realize we won't have a unique Lambda request context id, but we should at least discuss the pros/cons of the existing pattern we're using elsewhere.

`%d{yyyy-MM-dd HH:mm:ss.SSS} %X{AWSRequestId} %-5p %c{1} - %m%n`

PoeppingT:
Thanks for the callout @brtrvn, it makes sense to standardize our log4j2 configuration across the project. 

Here's a description of what our current (non-installer) log4j2 pattern really means:
`%d{yyyy-MM-dd HH:mm:ss.SSS} %X{AWSRequestId} %-5p %c{1} - %m%n`
* `%d{yyyy-MM-dd HH:mm:ss.SSS}`: the date, formatted as year-month-day hour:minute:second.millisecond
* `%X{AWSRequestId}`: as mentioned in the [AWS Lambda log4j2 documentation](https://docs.aws.amazon.com/lambda/latest/dg/java-logging.html#java-logging-log4j2) this will be filled with the Lambda request id
   * to me this is roughly equivalent to a "thread name". it feels superfluous to have it for the installer, but I'm happy to include it. if we ever introduce multithreading in the installer process we can consider changing our logging configurations to include that information as well
* `%-5p`: the log level (INFO, WARN, DEBUG, or ERROR) justified to a width of exactly 5 characters
* `%c{1}`: prints out the classname of the logger that published the event. `{1}` indicating that only the classname is included, none of the package information
   * as I've included in this PR, my preference is `%C{1}`, which instead prints the classname of the caller issuing the log request. with highly nested class structures it can sometimes make sense for an abstract parent class to hold the logger and implementing classes to use that logger, in which case it would be best for the actual classname issuing the log request (the implementer) to be listed, rather than the logger itself.
* `%m`: the actual log message
* `%n`: newline

My only suggestion to change that format is to use `%C` instead of `%c`, as I mention above.

Now that I'm looking more closely at the log4j2 configuration in the installer, we're including both `%M` and `%ex`, which are "the method name where the logging request was issued" and "the Throwable trace bound to the logging event" respectively. `%M` is described as an expensive operation, so my vote is we remove it. We will be able to find the log line without that. Similarly with `%ex`: from my testing it doesn't actually do anything useful, and might include stack trace information where we don't explicitly request it, which could clog up logs (and lead to a higher CW Logs cost), so I vote we remove that too.

My suggestion then is that we use exactly what's already provided, but `%C` instead of `%c`, leading to a final log4j2 configuration of `%d{yyyy-MM-dd HH:mm:ss.SSS} %X{AWSRequestId} %-5p %C{1} - %m%n` in all places, with the option to add in a "thread name" pattern (`%t`) once we decide it's necessary.

brtrvn:
Agree on the change from lowercase c to uppercase C.

Should we replace `%X{AWSRequestId}` in the installer log4j config with `%t` or does `%X{AWSRequestId}` properly no-op when the code is not running in a Lambda execution environment?

PoeppingT:
In a simple local test (changing the log4j2 configuration in the installer, running `build.sh` and immediately exiting) I'm seeing that `%X{AWSRequestId}` does just fill in as empty when there is no request id.

I like the idea of adding a thread name pattern, but it isn't useful right now and I don't see an immediate need for it, so I'm leaning that we leave it for later implementation.

brtrvn:
Let's keep our eye on confusion from users about AWS Request Id in the installer when it doesn't exist and also think about how we might encourage/enforce our standardized pattern string for new code.

Issue #90, opened 2021-06-29
File system settings for FSX with Windows workload does not get saved


4patelr:
I configured an application for Windows and when I try to configure the File system the settings are not being saved. 

### Reproduction Steps

1. Setup a Windows host in the Application page.
2. Deployed a tenant
3. Deleted a tenant
4. Tried to setup File system in the application page
5. Settings do not save

### Environment

  - **AWS Region  :** us-east-1
  - **AWS SaaS Boost Version :**
  - **Workload OS  (Linux or Windows) :**. Windows

### Other

--- 

This is :bug: Bug Report

4patelr:
The workaround for this is to toggle the O/S to Linux and then back to Windows on the Application config page before entering the file system details.

tobuck-aws:
The problem is that somehow the container OS was set without setting the `fileSystemType` on the config object. There's downstream logic that looks at `fileSystemType` to determine which configuration to send (either FSX or EFS), and because it was unset, neither was sent. I changed the code to default the `fileSystemType` to FSX if containerOS is initially set to Windows, and EFS if linux. Toggling OS after this should do the right thing as there's already a listener on the radio button that changes `fileSystemType` accordingly.

PR #91, opened 2021-06-29
Fix for #90

tobuck-aws:
Now defaults filesystemtype

Note that the changes seem significant, I accidentally left my prettier on 'format on change' so it just did a bunch of formatting. Really, the only significant change is on line 113 of ApplicationComponent.js.

Fix for #90 
----

### DIFF
--- a/client/web/src/settings/ApplicationComponent.js
+++ b/client/web/src/settings/ApplicationComponent.js
@@ -14,25 +14,21 @@
  * limitations under the License.
  */
 
-import React from "react";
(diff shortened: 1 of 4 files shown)


PoeppingT:
PR LGTM. When we work through Boost to reevaluate systems, I'd like for us to clean up this logic so we aren't switching so much on static strings. This makes sense as a hotfix.

tobuck-aws:
@PoeppingT 100% agree. Thanks for the review.

Issue #92, opened 2021-06-30
Tenant stack with FSX fails with missing lambdas folder after Boost upgrade


4patelr:
The FSX nested stack for tenant onboarding fails with missing key for the FsxDnsNameFunction.

### Reproduction Steps

1. Configure application for Windows with FSX
2. Update SaaS Boost using installer
2. A new lambda folder is created in the S3 artifact bucket
3. Onboard a new tenant
4. Tenant stack has "lambdas" as the Lambda Source Folder parameter value but that key no longer exists in the S3 bucket
5. FSX stack fails due to missing key for lambdas in
        S3Key: !Sub ${LambdaSourceFolder}/FsxDnsName-lambda.zip

### What did you expect to happen?
Tenant stack should use the newly created lambda-2021-06-30 folder for the source and not lambdas.  

Provision a tenant after update of SaaS Boost

### What actually happened?
Stack failed to create with missing key

### Environment

  - **AWS Region  :** eu-west-1
  - **AWS SaaS Boost Version :** 
  - **Workload OS  (Linux or Windows) :** Windows (and FSX)

### Other

--- 

This is :bug: Bug Report

4patelr:
Currently, running Update option from Installer creates the new folder with the date and uploads the lambda artifacts to it. The onboarding of tenants should pass that folder name for the LambdaSourceFolder instead of using the default lambdas.

4patelr:
Fixed by PR #95

PR #95, opened 2021-07-01
Save the lambdas source folder as a parameter

brtrvn:
Save the lambdas source folder as a parameter from the main SaaS Boost stack so it can be passed to the tenant onboarding stack(s). Resolves the immediate cause of Issue #92.

To make this work, the Onboarding service now needs to invoke the Settings service during provisioned tenant update to go fetch the necessary settings for the lambda folder and other CloudFormation data.

I also discovered during testing the CloudFormation considers a no updates to perform state to be an error. I'm now swallowing this error and just logging it.

There are some updates (for example just changing the lambdas folder) that _does not_ trigger a CloudFormation update. I could not figure out a way to force CloudFormation to perform a stack update consistently. I did add the metadata property to the onboarding templates as suggested by the documentation as a way to get CloudFormation to notice a change.

This flow also brings up some questions about whether the installer should be the place controlling these installation state changes and whether we should trigger a CloudFormation update on all provisioned tenants when we run update from the installer (we do not do that now). This could leave tenant environments in various states of infrastructure setup...

This PR also cleans up old developer cruft in the tenant onboarding CloudFormation template and clarifies that a tenant is either overriding the default compute settings or they're not. When override defaults is FALSE, the tenant does not have values for the individual compute settings.

----

### DIFF
--- a/resources/saas-boost-svc-onboarding.yaml
+++ b/resources/saas-boost-svc-onboarding.yaml
@@ -1180,6 +1180,9 @@ Resources:
       Environment:
         Variables:
           ONBOARDING_TABLE: !Ref OnboardingTable
+          API_TRUST_ROLE: !Sub arn:aws:iam::${AWS::AccountId}:role/sb-private-api-trust-role-${Environment}-${AWS::Region}
+          API_GATEWAY_HOST: !Sub ${SaaSBoostPrivateApi}.execute-api.${AWS::Region}.amazonaws.com
+          API_GATEWAY_STAGE: !Ref PrivateApiStage
       Tags:
         - Key: "Application"
           Value: "SaaSBoost"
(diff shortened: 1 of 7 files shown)


PoeppingT (approved):
Changes LGTM

4patelr:
Tested out the changes with a new install of SaaS Boost. Deployed a tenant and that points to the lambdas folder. Then ran the update using installer which created the new Lambda source folder. After which I onboarded another tenant and the lambda source folder in the tenant (and nested stacks) pointed to the new folder so that part looks good.

I tested the update of a tenant by changing the compute for the first provisioned tenant and this triggered the CloudFormation update on the main tenant stack. However, no update was triggered for the nested RDS and EFS stacks and the LambdaSourceFolder parameter for the RDS stack remained as lambdas. I think this aligns to what you found with how CloudFormation works.

4patelr (approved):
Changes look good.

PR #104, opened 2021-07-07
Use general purpose SSD storage by default instead of magnetic for RDS instances

brtrvn:
RDS suggests all new database instances be created with general purpose SSD storage for better performance and access to additional features like storage auto-scaling. CloudFormation's default is to use magnetic storage.

Use `gp2` for the RDS Instance storage type.

----


brtrvn:
See discussion #102

4patelr (approved):
Look good to me.

Issue #70, opened 2021-06-11
Stop the services when not in use


adilbasha:
Stop the SaaS boost services when not in use

### Use Case

From a Dev project standpoint, we use the environment during working hours. we would like to see the feature in stopping the service in nonworking works.
For now, in fargate we are changing the autoscaling group to min 0 desired 0 and max as 0

### Proposed Solution

In the SaaS boast web console, under each tenant have the option to change the fargate ASG and also have the option to launch in the public subnet to avoid TSG cost

### Other

* [ ] :wave: I may be able to implement this feature request
* [ ] :warning: This feature might incur a breaking change

---

This is a :rocket: Feature Request

mahara13:
I am also interested in this functionality

brtrvn:
Resolved with PR #109 
Be sure to re-run the installer after merging and choose `Update existing AWS SaaS Boost deployment`

PR #109, opened 2021-07-17
Shutdown and Startup tenant ECS services for non-production environments

brtrvn:
Implementation for Feature Request #70. Optional functions to stop and start all provisioned tenant ECS services running the application workload in non-production environments. The Lambda functions and supporting log groups and IAM permissions will be installed with SaaS Boost, but no trigger mechanism will be automatically installed.

We don't want to encourage the use of this functionality in production SaaS environments, but recognize the cost savings for non-production environments that do not need to be available 24/7 could be significant. See the enable/disable scripts in each directory for examples of how to use EventBridge scheduled events to trigger these functions.


functions/ecs-shutdown-services/enable.sh
```diff
@@ -0,0 +1,103 @@
...
+    SCHEDULE="${EXISTING_SCHEDULE}"
+fi
+
+# Simple pattern check -- doesn't test for EventBridge restriction on
+# day-of-month and day-of-week not both being asterisks
+CRON_REGEX="^cron\(.+ .+ .+ .+ .+ .+\)$"
+if ! [[ $SCHEDULE =~ $CRON_REGEX ]]; then
+    echo "Schedule cron expression ${SCHEDULE} is not valid."
+    exit 1
+fi
```
PoeppingT, line 63: What if I don't want to set up a schedule for the EventBridge rule? Should there be a way to just use this scripts to fire the event for the disable/enable services lambdas?

functions/ecs-shutdown-services/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/EcsShutdownServices.java
```diff
@@ -0,0 +1,135 @@
...
+                    LOGGER.error(Utils.getFullStackTrace(ecsError));
+                    throw ecsError;
+                }
+            }
+        }
+
+        return null;
+    }
+
+    public ArrayList<Map<String, Object>> getProvisionedTenants(Context context) {
```
PoeppingT, line 112: Just an aside, this should really be extrapolated into a TenantsClient object into a layer. The ECSShutdownServices Lambda shouldn't need to know the static path required to get the provisioned tenants.

PoeppingT: nit: `s/public/private/`

functions/ecs-shutdown-services/README.md
```diff
@@ -0,0 +1,16 @@
+# ECS Shutdown Services
+
+## Overview
+This function can be used to gracefully shutdown all application workloads running for your provisioned tenants. It does this by setting the `desiredCount` attribute of the ECS service for each tenant to zero (0). [Watch a detailed walkthru](https://www.twitch.tv/videos/1065389231) of building this solution during our [Office Hours](https://github.com/awslabs/aws-saas-boost/discussions/106).
+
+## Why would you turn off your SaaS application?
+Excellent question! You should **not** use this function for your production environments. Your SaaS customers expect your service to be available at all times. However, for development and other non-production environments, it may be useful to temporarily shutdown your application tasks as a way to save operational costs. ECS tasks that run on Fargate are only billed when they are running. If you are running your tasks on EC2, you will still incur the costs of the EC2 instances in your ECS cluster.
+
+## How do I use it?
```
PoeppingT, line 9: The example `enable.sh` and `disable.sh` scripts fail if the lambdas have not been set up. Should the READMEs include some instructions for updating an existing SaaS Boost installation to include these Lambdas?

functions/ecs-shutdown-services/enable.sh
brtrvn: Why wouldn't you just invoke the Lambda if you didn't want it on a schedule? You don't need a script for that.
`aws lambda invoke --function-name sb-${SAAS_BOOST_ENV}-ecs-shutdown-services response.json`

functions/ecs-shutdown-services/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/EcsShutdownServices.java
brtrvn: Not sure I agree with that -- if I'm going to call the EC2 REST API, I have to know the path to the resource I want to invoke. This isn't an SDK, it's just a REST API.

brtrvn: I made it public for unit test visibility (which I didn't get around to yet). Happy to make it protected or package private.

functions/ecs-shutdown-services/README.md
brtrvn: Fair question. We don't include specific instructions on how to pull/merge with any other commits. The pattern is always to run the installer and choose update existing install. That will take care of everything.

brtrvn: And, to be clear, do the shell scripts crash or do they give you sensible output and exit?

functions/ecs-shutdown-services/enable.sh
PoeppingT: My understanding was that this PR is targeting ease of use improvements for developers of SaaS Boost. I personally wouldn't see having to manually invoke a lambda function as a particularly usable way to shutdown tenant services, especially if there is already a script that is basically doing the same thing.

I'll leave this up to you. If doing it without a schedule causes the script to fail, maybe some error message in the script would be useful:
``"Schedule cron expression ${SCHEDULE} is not valid. To just disable tenant services, run `aws lambda invoke --function-name sb-${SAAS_BOOST_ENV}-ecs-shutdown-services response.json`"``

functions/ecs-shutdown-services/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/EcsShutdownServices.java
PoeppingT: I'm simply suggesting that we should eventually write a client abstraction layer to the REST API, so we can safely make changes to the services (server in the client-server model) and versioned changes to the client without breaking existing installations. Each module should only know the direct information it **absolutely has to know**, treating all else as black boxes.

functions/ecs-shutdown-services/README.md
PoeppingT: Once again, maybe the best option here is to suggest running the installer if we find that the lambdas haven't been installed. If we can lower the bar for entry for a new contributor, we should.

4patelr (approved):
LGTM. Ready for merge

functions/ecs-startup-services/enable.sh
```diff
@@ -0,0 +1,103 @@
...
+
+RULE="sb-${SAAS_BOOST_ENV}-ecs-startup-services"
+EXISTING_SCHEDULE=$(aws events describe-rule --name "${RULE}" --query "ScheduleExpression" --output text)
+if [ $? == 0 ] && [ ! -z "${EXISTING_SCHEDULE}" ]; then
+    read -p "Reuse the existing schedule expression ${EXISTING_SCHEDULE}? [Y/N] " REUSE_SCHEDULE
+fi
+if ! [[ $REUSE_SCHEDULE =~ ^[Yy] ]]; then
+    # Get a cron schedule to invoke our Lambda on
+    echo "Enter an EventBridge cron schedule expression (https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-create-rule-schedule.html#eb-cron-expressions)."
+    read -p "[Press enter for default schedule of 6 AM daily]: " SCHEDULE
```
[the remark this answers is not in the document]
brtrvn: What exactly do you envision it looking like? We're already pointing them at the documentation (because EventBridge _almost_ but not quite follows the cron syntax standard). We're also letting them just hit the enter/return key to accept a default.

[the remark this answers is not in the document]
4patelr: Something simple like making the prompt as follows so user can change the time very easily.
[Press enter for default schedule of 6 AM daily, (0 6 * * ? *)]

functions/ecs-shutdown-services/enable.sh
```diff
@@ -0,0 +1,103 @@
...
+
+RULE="sb-${SAAS_BOOST_ENV}-ecs-shutdown-services"
+EXISTING_SCHEDULE=$(aws events describe-rule --name "${RULE}" --query "ScheduleExpression" --output text)
+if [ $? == 0 ] && [ ! -z "${EXISTING_SCHEDULE}" ]; then
+    read -p "Reuse the existing schedule expression ${EXISTING_SCHEDULE}? [Y/N] " REUSE_SCHEDULE
+fi
+if ! [[ $REUSE_SCHEDULE =~ ^[Yy] ]]; then
+    # Get a cron schedule to invoke our Lambda on
+    echo "Enter an EventBridge cron schedule expression (https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-create-rule-schedule.html#eb-cron-expressions)."
+    read -p "[Press enter for default schedule of 12 midnight daily]: " SCHEDULE
```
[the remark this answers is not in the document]
4patelr: A prompt like:
[Press enter for default schedule of 12 midnight daily, (0 0 * * ? *)]

PoeppingT (approved):

PR #110, opened 2021-07-20
Use existing REDSHIFT_MASTER_PASSWORD SSM param, if present

4patelr:
The installer would fail on install of Metrics and Analytics if an environment was previous deleted because the REDSHIFT_MASTER_PASSWORD SSM parameter would remain and a new version would created. This PR will make use of existing version of the parameter if found.

### DIFF
--- a/installer/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/SaaSBoostInstall.java
+++ b/installer/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/SaaSBoostInstall.java
@@ -380,23 +380,29 @@ public class SaaSBoostInstall {
                 .build();
         buildAndCopyLambdas("resources" + File.separator + "custom-resources", list);
 
-        //create db password if metrics installed
-        if (null == dbPassword || dbPassword.isEmpty()) {
-            UUID uniqueId = UUID.randomUUID();
-            String[] parts = uniqueId.toString().split("-");  //UUID 29219402-d9e2-4727-afec-2cd61f54fa8f
-            dbPassword = "Pass12" + parts[0];
-            //outputMessage("The database password for Redshift Metrics database is: " + this.dbPassword);
-        }
-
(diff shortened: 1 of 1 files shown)


installer/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/SaaSBoostInstall.java
```diff
@@ -380,23 +380,29 @@ private void installMetrics(String sbStackName) throws IOException {
...
+                    .build());
+            dbPassword = parameterResponse.parameter().value();
+            LOGGER.info("Found existing SSM Parameter: " + dbPasswordSSMParameter);
+        } catch (Exception e) {
+            //create db password param if not existing
+            LOGGER.info("SSM Parameter " + dbPasswordSSMParameter + " not found, add it.");
+            if (null == dbPassword || dbPassword.isEmpty()) {
+                UUID uniqueId = UUID.randomUUID();
+                String[] parts = uniqueId.toString().split("-");  //UUID 29219402-d9e2-4727-afec-2cd61f54fa8f
+                dbPassword = "Pass12" + parts[0];
```
PoeppingT, line 401: It looks like this "Pass12" was added to guarantee meeting [Redshift's password requirements](https://docs.aws.amazon.com/redshift/latest/dg/r_CREATE_USER.html#r_CREATE_USER-parameters):

> * It must be 8 to 64 characters in length.
> * It must contain at least one uppercase letter, one lowercase letter, and one number.
> * It can use any ASCII characters with ASCII codes 33–126, except ' (single quotation mark), " (double quotation mark), \, /, or @.
Can you add a comment here to explain why a constant value is prepended to all passwords? Also, since UUID is always 36 characters, why not use the whole thing? This would increase the complexity of passwords, making them less vulnerable to attacks.

PoeppingT (commented):
Overall LGTM

Issue #64, opened 2021-06-04
Delete of AWS SaaS Boost environment does not delete the SSM parameters for the environment 


4patelr:
The SSM parameters created for the environment are not deleted when the environment is deleted using the installer or by deleting the CloudFormation stack.

### Reproduction Steps

1. Use installer to create a SaaS Boost environment named dev.
2. Go to AWS console and review the SSM parameters created for the environment with name. For example, /saas-boost/dev/DB_PORT
3. Use the installer to delete the SaaS Boost environment.

### What did you expect to happen?
Expected the SSM parameters for the environment to be deleted.

### What actually happened?
The SSM parameters such as /saas-boost/dev/DB_PORT are not deleted.

### Environment

  - **AWS Region  :** us-east-1
  - **AWS SaaS Boost Version :**
  - **Workload OS  (Linux or Windows) :** Linux

### Other

--- 

This is :bug: Bug Report

brtrvn:
Also make sure we clean up tenant parameters when we delete tenant from the Tenant Service.

PR #115, opened 2021-08-16
App config delete and installer cleanup

brtrvn:
These changes update the Settings service to be able to delete the application config from an existing install. This functionality is exposed through the private API.

The installer has also been reworked to clean up the code and be sure we're using the v2 SDK clients for all services as well as the preferred Java patterns for file system, dates and times, and stream processing.

The installer now invokes the new delete app config functionality as part of deleting a SaaS Boost install, resolving Issue #64. 

The installer has also been updated to improve how the analytics module is installed and updated, resolving Issue #73 and supersedes PR #110.

----


installer/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/SaaSBoostInstall.java
```diff
@@ -15,1678 +15,1109 @@
...
                 }
             } else {
-                outputMessage("Entered value for email address is incorrect or wrong format, please try again.");
-            }
-        } while (!isValid);
-
-        String useDomain = "n";
-        isValid = false;
-        do {
-            System.out.print("Would you like to setup a domain in Route 53 as a Hosted Zone for the AWS SaaS Boost environment (y or n)? ");
```
PoeppingT, line 1285: Why are we removing the ability to set up a Route53 Hosted Zone Domain for the newly installed environment?

services/settings-service/src/test/java/com/amazon/aws/partners/saasfactory/saasboost/SettingsServiceTest.java
```diff
@@ -254,6 +254,13 @@ public void testToParameterStore() {
     public void testFromParameterStore() {
 
     }
+
+    @Test
+    public void testToSettings() {
+        AppConfig appConfig = AppConfig.builder().build();
+        List<Setting> settings = SettingsServiceDAL.toSettings(appConfig);
+        System.out.println(Utils.toJson(settings));
+    }
```
PoeppingT, line 263: This doesn't look to actually verify the result of the toSettings() function. Why not add keys and values of various types to the AppConfig, then verify they exist in the Setting list returned by SettingsServiceDAL?

services/settings-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/SettingsServiceDAL.java
```diff
@@ -647,7 +714,10 @@ public static Parameter toTenantParameterStore(UUID tenantId, Setting setting) {
         settings.add(Setting.builder().name("MIN_COUNT").value(minCount).readOnly(false).build());
         String maxCount = appConfig.getMaxCount() != null ? appConfig.getMaxCount().toString() : null;
         settings.add(Setting.builder().name("MAX_COUNT").value(maxCount).readOnly(false).build());
-        settings.add(Setting.builder().name("CLUSTER_OS").value(appConfig.getOperatingSystem().name()).readOnly(false).build());
+        OperatingSystem os = appConfig.getOperatingSystem();
+        if (os != null) {
+            settings.add(Setting.builder().name("CLUSTER_OS").value(os.name()).readOnly(false).build());
+        }
```
PoeppingT, line 720: This will cause a null pointer in [tenant onboarding](https://github.com/awslabs/aws-saas-boost/blob/main/services/onboarding-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/OnboardingService.java#L765) if the `CLUSTER_OS` parameter isn't included.

services/settings-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/SettingsServiceDAL.java
```diff
@@ -546,6 +562,57 @@ public AppConfig getAppConfig() {
...
+                        break;
+                    }
+                } else {
+                    // Batch has reached max size of 10, make the request
+                    DeleteParametersResponse response = ssm.deleteParameters(r ->
+                            r.names(batch)
+                    );
+                    LOGGER.info("SettingsServiceDAL::deleteAppConfig removed " + response.deletedParameters().toString());
+                    if (response.hasInvalidParameters() && !response.invalidParameters().isEmpty()) {
+                        LOGGER.warn("SettingsServiceDAL::deleteAppConfig invalid parameters " + response.invalidParameters().toString());
```
PoeppingT, line 599: What does it mean about the state of the system if we reach this? Is there some way we can recover the failed deletion?

services/settings-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/SettingsServiceDAL.java
```diff
@@ -330,6 +334,18 @@ private Parameter putParameter(Parameter parameter) {
...
 
+    private void deleteParameter(Parameter parameter) {
+        try {
+            DeleteParameterResponse response = ssm.deleteParameter(request -> request
+                    .name(parameter.name())
+            );
+        } catch (SdkServiceException ssmError) {
+            LOGGER.error("ssm:DeleteParameter error " + ssmError.getMessage());
+            throw ssmError;
+        }
```
PoeppingT, line 345: This could be a [ParameterNotFoundException](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/ssm/model/ParameterNotFoundException.html). Does it make sense to throw the error upwards in this case? Trying to delete a Parameter but not finding it means it doesn't exist, so the operation should be successful, right?

services/settings-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/SettingsServiceDAL.java
```diff
@@ -308,6 +308,10 @@ public Setting updateSetting(Setting setting) {
         return fromParameterStore(updated);
     }
 
+    public void deleteSetting(Setting setting) {
+        deleteParameter(toParameterStore(setting));
+    }
```
PoeppingT, line 313: Why are these functions added, for future convenience? They don't appear to be used.

resources/saas-boost-private-api.yaml
```diff
@@ -944,6 +953,38 @@ Resources:
             method.response.header.Access-Control-Allow-Origin: false
             method.response.header.Access-Control-Max-Age: false
             method.response.header.X-Requested-With: false
+
+
+
```
PoeppingT, line 958: extra unnecessary whitespace?

installer/src/test/java/com/amazon/aws/partners/saasfactory/saasboost/CloudFormationTemplateParserTest.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.Arrays;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import static org.junit.Assert.*;
+
+public class CloudFormationTemplateParserTest {
+
+    @Test
+    public void testReadTemplate() throws Exception {
```
PoeppingT, line 32: Currently this tests the jackson YAMLFactory. Why not test installers `getCloudFormationParameterMap()`?

PoeppingT (commented):
Here's an initial review. Lots of changes in the installer, so I'll keep working on it in the coming days. I wanted to get a first review out.

installer/pom.xml
```diff
@@ -76,9 +74,11 @@ limitations under the License.
                         <configuration>
                             <shadedArtifactAttached>true</shadedArtifactAttached>
                             <transformers>
-                                <transformer implementation=
-                                                     "org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
+                                <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
                                     <mainClass>com.amazon.aws.partners.saasfactory.saasboost.SaaSBoostInstall</mainClass>
+                                    <manifestEntries>
+                                        <Multi-Release>true</Multi-Release>
+                                    </manifestEntries>
```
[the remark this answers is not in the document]
brtrvn: It gets rid of this scary sounding warning that prints to STDOUT when you execute the shaded JAR. It's an issue with the shade maven plugin.
WARNING: sun.reflect.Reflection.getCallerClass is not supported. This will impact performance.

installer/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/Keyboard.java
```diff
@@ -0,0 +1,122 @@
...
+
+    public static String readWord() {
+        String value = null;
+        try {
+            value = reader.next().split("\\s+")[0];
+        } catch (Exception exception) {
+            LOGGER.error("Error reading String data, null value returned.");
+        }
+        return value;
+    }
```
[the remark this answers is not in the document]
brtrvn: Wouldn't that be public static String[] readWords() ?
We can remove it.

installer/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/Keyboard.java
```diff
@@ -0,0 +1,122 @@
...
+
+public class Keyboard {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(Keyboard.class);
+    private static Scanner reader = new Scanner(System.in);
+    static {
+        reader.useDelimiter(System.lineSeparator());
+    }
+
+    public static String readString() {
```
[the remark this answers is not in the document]
brtrvn: Good idea

services/settings-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/SettingsServiceDAL.java
brtrvn: It seemed like a warning vs. an exception to me. It means you probably sent in parameter names that don't exist. The documentation says this collection will be non empty when there are "invalid" parameters, but does not define what invalid means. We log it so that people can have an audit trail if things go wrong. This code branch is a bit pedantic, since the list of parameters to delete is fetch immediately prior to this and I wouldn't expect `getParameters` to return invalid parameters.

services/settings-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/SettingsServiceDAL.java
brtrvn: Good design question. Do you want to know that you're trying to delete something that doesn't exist, or should that just be a warning (and swallow the `ParameterNotFoundException`)? I'm ok either way.

services/settings-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/SettingsServiceDAL.java
brtrvn: They were added as I started working on the functionality and I switched gears to using the batch delete method from the SDK instead. I can see future functionality coming from them, but you are correct, they are not used today.

installer/src/test/java/com/amazon/aws/partners/saasfactory/saasboost/CloudFormationTemplateParserTest.java
brtrvn: That was the original intent, but torn on what to do about the method arguments. Do you suggest that the caller of `getCloudFormationParameterMap` own resolving the `java.nio.file.Path` of the template file and passing it in as an argument? It's either that or the test class would have to set `workingDir` which does not currently expose a traditional setter.

resources/saas-boost-private-api.yaml
brtrvn: Oops

Issue #47, opened 2021-05-25
Changing billing plans creates a new Customer object in Stripe


aspcanada:
Changing billing plans creates a new customer object and subscription in Stripe.

### Reproduction Steps

Select tenant > click edit > change billing plan

### What did you expect to happen?

Existing Stripe subscription is updated to reflect new plan and prorated billing shows up in a new invoice for the next month.

### What actually happened?

A new Customer is created with a new subscription. Tenant is charged multiple times.

### Environment

  - **AWS Region  :** us-west-2
  - **AWS SaaS Boost Version :** v0
  - **Workload OS  (Linux or Windows) :** Linux

### Other

--- 

This is :bug: Bug Report

PR #116, opened 2021-08-17
Fix for change of billing plan

4patelr:
Fixes Issue 47 by canceling existing subscription and creating a new subscription in Stripe when the billing plan for a Tenant is updated.

### DIFF
--- a/metering-billing/lambdas/src/main/java/com/amazon/aws/partners/saasfactory/metering/onboarding/BillingIntegration.java
+++ b/metering-billing/lambdas/src/main/java/com/amazon/aws/partners/saasfactory/metering/onboarding/BillingIntegration.java
@@ -303,7 +303,7 @@ public class BillingIntegration implements RequestHandler<EventBridgeEvent, Obje
                                 .build();
                 customer = Customer.create(params);
                 //create the subscription
-                createStripeSubscription(customer, planEnum, tenantId,false);
+                createStripeSubscription(customer, planEnum, tenantId);
                 LOGGER.info("provisionTenantInStripe: Customer created with id: " + customer.getId());
             } catch (StripeException e) {
                 LOGGER.error("provisionTenantInStripe: Error creating Customer for Tenant: " + tenantId);
@@ -351,7 +351,7 @@ public class BillingIntegration implements RequestHandler<EventBridgeEvent, Obje
                 return;
             }
 
-            createStripeSubscription(customer, planEnum, tenantId,true);
+            createStripeSubscription(customer, planEnum, tenantId);
         }
     }
 
@@ -428,7 +428,7 @@ public class BillingIntegration implements RequestHandler<EventBridgeEvent, Obje
     For the metered products, a message is sent to Event Bus to add that record to the Dynanamo table mapping
      the Tenant internal product code to the Stripe Subscription id which is unique for each tenant and product.
      */
-    private void createStripeSubscription(Customer customer, SubscriptionPlan plan, String fullTenantId, boolean existingCustomer) throws StripeException{
+    private void createStripeSubscription(Customer customer, SubscriptionPlan plan, String fullTenantId) throws StripeException{
         //get the PRICE ID for the product code of the plan.  This is for the monthly subscription
         final String planPriceId = getPriceId(plan.name());
         List<SubscriptionCreateParams.Item> items = new ArrayList<>();
--- a/metering-billing/lambdas/src/main/java/com/amazon/aws/partners/saasfactory/metering/onboarding/BillingIntegration.java
+++ b/metering-billing/lambdas/src/main/java/com/amazon/aws/partners/saasfactory/metering/onboarding/BillingIntegration.java
@@ -24,6 +24,7 @@ import com.amazon.aws.partners.saasfactory.saasboost.Utils;
 import com.amazonaws.services.lambda.runtime.Context;
 import com.amazonaws.services.lambda.runtime.RequestHandler;
 import com.stripe.Stripe;
+import com.stripe.exception.InvalidRequestException;
 import com.stripe.exception.StripeException;
 import com.stripe.model.*;
 import com.stripe.net.StripeResponse;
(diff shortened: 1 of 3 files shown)


4patelr:
#47 - Fixes issue 47

metering-billing/lambdas/src/main/java/com/amazon/aws/partners/saasfactory/metering/onboarding/BillingIntegration.java
```diff
@@ -271,58 +272,94 @@ private void provisionTenantInStripe(String tenantId, String planId) throws Stri
...
+                LOGGER.info("provisionTenantInStripe: Customer created with id: " + customer.getId());
+            } catch (StripeException e) {
+                LOGGER.error("provisionTenantInStripe: Error creating Customer for Tenant: " + tenantId);
+                LOGGER.error(Utils.getFullStackTrace(e));
+                throw e;
+            }
+        } else {
+            LOGGER.info("provisionTenantInStripe: Customer: {} create skipped, already exists with id: {} ", customerName, customer.getId());
+            //If subscription plan has changed then cancel the old subscriptions and create a new.
+            //Ideally, the subscription information for the tenant would be stored in a DDB table so we do not have to hit Stripe to find if it has changed
```
PoeppingT, line 316: If the customer can change subscription information directly in Stripe, we should *not* store any subscription info in DDB as that information may become inconsistent. We can build optimizations around Stripe (or other billing providers) if we find that it becomes a bottleneck.

metering-billing/lambdas/src/main/java/com/amazon/aws/partners/saasfactory/metering/onboarding/BillingIntegration.java
```diff
@@ -271,58 +272,94 @@ private void provisionTenantInStripe(String tenantId, String planId) throws Stri
...
-        LOGGER.info("provisionTenantInStripe: Setting up Tenant: " + customerId + " as a customer and creating subscription");
+        final String customerName = "tenant-" + tenantId.split("-")[0];  //get first segment
+        LOGGER.info("provisionTenantInStripe: Setting up Tenant: {} as a customer and creating subscription", customerName);
 
-        try {
-            //Only attempt to setup tenant if customer does not already exist for this tenant
-            try {
-                Customer customer = Customer.retrieve(customerId);
-                LOGGER.info("Customer : " + customerId + " already exists, setup of tenant: " + tenantId + " skipped");
+        Customer customer = getStripeCustomer(customerName);
```
PoeppingT, line 283: What happens if Stripe is down or times out when responding? Do we fail the entire tenant onboarding stack? Should we retry, either immediately or lazily?

metering-billing/lambdas/src/main/java/com/amazon/aws/partners/saasfactory/metering/onboarding/BillingIntegration.java
```diff
@@ -271,58 +272,94 @@ private void provisionTenantInStripe(String tenantId, String planId) throws Stri
...
+                            LOGGER.info("Subscription Plan has not changed for the tenant, nothing to do");
+                            return;
+                        }
+                    }
+                    //NOTE: Do queued metered events need to sent to Stripe before canceling the subscription?
+
+                    //attempt to cancel the subscription
+                    LOGGER.info("provisionTenantInStripe: No match in Stripe subscription items for plan {}", planId);
+                    LOGGER.info("provisionTenantInStripe: Attempt to Cancel Subscription: {} ", subscription.getId());
+                    Subscription canceledSubscription = subscription.cancel();
```
PoeppingT, line 337: Why not [update the subscription](https://stripe.com/docs/api/subscriptions/update) instead of cancelling and recreating?

metering-billing/lambdas/src/main/java/com/amazon/aws/partners/saasfactory/metering/onboarding/BillingIntegration.java
```diff
@@ -366,25 +388,47 @@ private void cancelSubscriptionInStripe(String tenantId) throws StripeException
...
 
     /*
-    Creates a customer and a subscription in Stripe
+    Creates a subscription in Stripe
     The Subscription will have one item for the monthly plan and additional items for metered products
     For the metered products, a message is sent to Event Bus to add that record to the Dynanamo table mapping
      the Tenant internal product code to the Stripe Subscription id which is unique for each tenant and product.
      */
-    private void createStripeSubscription(Customer customer, SubscriptionPlan plan, String fullTenantId) throws StripeException{
+    private void createStripeSubscription(Customer customer, SubscriptionPlan plan, String fullTenantId, boolean existingCustomer) throws StripeException{
```
PoeppingT, line 431: We're now passing in `existingCustomer` but it doesn't appear you're using it.

resources/custom-resources/disable-instance-protection/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/DisableInstanceProtection.java
```diff
@@ -55,8 +55,8 @@ public Object handleRequest(Map<String, Object> event, Context context) {
         LOGGER.info("Processing for Autoscaling group {}", autoScalingGroup);
         try {
             Runnable r = () -> {
-                if ("Delete".equalsIgnoreCase(requestType)) {
-                    LOGGER.info("DELETE");
+                if ("Delete".equalsIgnoreCase(requestType) || "Update".equalsIgnoreCase(requestType)) {
+                    LOGGER.info("Request Type: " + requestType);
```
PoeppingT, line 59: These changes don't appear to be related to fixing billing. Can you explain why this and the change to saas-boost-svc-onboarding.yaml are in the diff?

metering-billing/lambdas/src/main/java/com/amazon/aws/partners/saasfactory/metering/onboarding/BillingIntegration.java
PoeppingT: In this case by customer I meant the SaaS vendor/builder/ISV using Boost, as I would assume they have the highest level of control eventually over the Stripe subscription of behalf of the tenant; after all, the Stripe subscription is so the ISV can bill the tenant for usage.

I'm fine with sticking with this current implementation, I'm just suggesting storing information in DDB may have some unforeseen side effects.

metering-billing/lambdas/src/main/java/com/amazon/aws/partners/saasfactory/metering/onboarding/BillingIntegration.java
PoeppingT: Both great questions. Sounds like when we pick up this PR we'll need to do some experimentation with Stripe's API to see how it acts in different situations.

metering-billing/lambdas/src/main/java/com/amazon/aws/partners/saasfactory/metering/onboarding/BillingIntegration.java
brtrvn: Agree that we should have some sort of async retry mechanism in place for 3rd party integrations, but we are currently safe from blowing up the entire tenant onboarding/tenant update flow since we invoke these billing service api's async currently.

resources/custom-resources/disable-instance-protection/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/DisableInstanceProtection.java
4patelr: The current flow is when a tenant is updated in UI, it triggers a CloudFormation Stack Update. This change is necessary to send the message to the BillingService with the Billing Plan in case the Billing Plan has changed. The Billing Service will determine if the plan has changed.

metering-billing/lambdas/src/main/java/com/amazon/aws/partners/saasfactory/metering/onboarding/BillingIntegration.java
4patelr: Upon further research, Stripe does allow updating the subscription with a new plan and it will do the Proration. 
Take a look at https://stripe.com/docs/billing/subscriptions/change and https://stripe.com/docs/billing/subscriptions/prorations. May be we need to look at this approach instead of the cancel and create new subscription.

metering-billing/lambdas/src/main/java/com/amazon/aws/partners/saasfactory/metering/onboarding/BillingIntegration.java
4patelr: I thought if the Lambda errors out the message from EventBridge will be retried.

metering-billing/lambdas/src/main/java/com/amazon/aws/partners/saasfactory/metering/onboarding/BillingIntegration.java
4patelr: Updated to remove unused variable

PR #118, opened 2021-08-31
Reorganize Maven modules + add checkstyle/spotbugs

PoeppingT:
This commit contains changes to prepare aws-saas-boost for CI automation via GitHub Actions. Primarily, this commit reorganizes the Maven module structure of the project to be one based on pom.xml in the project's root directory. This means that inter-module dependencies are taken care of automatically, as well as giving users an easy way to build the project with one command, rather than manually module-by-module. 

This commit also includes the following additional changes:
* remove the "local-repo" workaround to enable individual module building
* add maven-checkstyle (and a copy of [google-checks.xml](https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml) with increased line length) to the repository
* configure checkstyle in each existing module to fail the build if any new style violations are introduced
* add maven-spotbugs to the repository
* standardize the maven-compile, maven-surefire, and maven-assembly plugins across multiple modules
* remove the unused maven-source plugin from the layers/utils module
* other miscellaneous changes

----


brtrvn:
Overall comments - this solution supports building the entire solution (other than the installer) from the root level directory, but does not support building from middle level directories. For example, I can build everything from `/pom.xml`, but I can't build all the services from `/services/pom.xml`. I can still build a service individually from its pom file, but only if I build the projects in the `/layers` directory first. Note that this limitation exists today (which is why we have shell scripts to build the various lambdas).

Due to the parameterization of some of the dependency versions, have we broken the ability to run a single artifact's pom without running the parent/grandparent pom?

The checkstyle error count seems arbitrary, but I understand there are some getting started heuristics employed to give us a baseline. I'd be concerned about supporting so many different values going forward. Maybe there's a better way in the future?

PoeppingT:
> Overall comments - this solution supports building the entire solution (other than the installer) from the root level directory, but does not support building from middle level directories. For example, I can build everything from `/pom.xml`, but I can't build all the services from `/services/pom.xml`. I can still build a service individually from its pom file, but only if I build the projects in the `/layers` directory first. Note that this limitation exists today (which is why we have shell scripts to build the various lambdas).

This is true, and is how maven is intended to work. If you want to build against the changes you've made to the `layers/` locally, of course you need to build, right? Otherwise how could maven possibly figure things out?

In the case that you're just trying to build against the latest version of SaaS Boost without having to build `layers/` manually, this is why we need a release process. If we publish artifacts to maven central (or any other repository) then those dependencies can be downloaded automatically instead of needing to be build locally.

As a case study, take [Hadoop](https://github.com/apache/hadoop). If you clone `mainline` Hadoop, delete everything in your local maven repository, and try to build `hadoop-hdfs-project/hadoop-hdfs/`, it will fail because the SNAPSHOT versions are not published to any external repository and you have not built those dependencies locally. However, if you switch to the 3.3.0 release (using `git checkout rel/release-3.3.0`) and try building again, it will succeed because the dependencies have been downloaded from Maven central.

Switching this paradigm to one of downloading artifacts will be a difficult one (because the installer currently makes some assumptions about the project structure) but will bring us closer to how other open source projects typically do things.

> Due to the parameterization of some of the dependency versions, have we broken the ability to run a single artifact's pom without running the parent/grandparent pom?

Nope, I just tested with `layers/`. Even after deleting all local Maven repository information and rebuilding, tests + checkstyle are still able to run correctly by reading the parents, even if they're not directly invoked. This is why the parent pom is directly referenced in each pomfile, creating a dependency tree.

> The checkstyle error count seems arbitrary, but I understand there are some getting started heuristics employed to give us a baseline. I'd be concerned about supporting so many different values going forward. Maybe there's a better way in the future?

Ideally we get all the checkstyle errors down to zero, and as part of building unit tests and refactoring existing code I will work to accomplish that. Then we don't need overrides in each module and can just rely on the default declared in the parent pomfile.

brtrvn (approved):

PR #125, opened 2021-09-14
Check for number of errors from ECR before failing

### LINKED ISSUE #124: Deleting SaaS Boost environment from Installer fails with ECR error
When you delete a SaaS Boost environment using the installer, it removes the docker images from the SaaS Boost ECR repository before running CloudFormation delete stack because CloudFormation cannot remove an ECR repo that isn't empty. The AWS SDK libraries commonly have a "_hasErrors_" or "_hasFailures_" type boolean method on the various response objects, but the _hasXXX_ methods return true when the underlying `java.util.Collection` is not null even if the collection is empty. The installer is throwing a runtime exception every time because it's only checking for hasFailures and not also looking at how many failures exist.

### Reproduction Steps
1. Install SaaS Boost
2. Push a docker image to the SaaS Boost ECR repo
3. Delete SaaS Boost from the installer script

The SaaS Boost ECR repo to be emptied and the SaaS Boost installation to be deleted

### What actually happened?
The installer program exits prematurely with
```
INFO  SaaSBoostInstall - Installation Error: ECR delete image failures 0
INFO  SaaSBoostInstall - Please see detailed log file saas-boost-install.log
ERROR SaaSBoostInstall - java.lang.RuntimeException: ECR delete image failures 0
    at com.amazon.aws.partners.saasfactory.saasboost.SaaSBoostInstall.deleteEcrImages(SaaSBoostInstall.java:989)
```

### Environment

  - **AWS Region  :**

### Other

--- 

This is :bug: Bug Report

brtrvn:
Fixes #124 

Check for both existence of failures collection and size greater than zero before exiting installer with exception.

----

### DIFF
--- a/installer/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/SaaSBoostInstall.java
+++ b/installer/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/SaaSBoostInstall.java
@@ -983,7 +983,7 @@ public class SaaSBoostInstall {
                 if (response.hasImageIds()) {
                     LOGGER.info("Deleted " + response.imageIds().size() + " images");
                 }
-                if (response.hasFailures()) {
+                if (response.hasFailures() && !response.failures().isEmpty()) {
                     LOGGER.error("Error deleting images from ECR");
                     response.failures().forEach(failure -> LOGGER.error("{} {}", failure.failureCodeAsString(), failure.failureReason()));
                     throw new RuntimeException("ECR delete image failures " + response.failures().size());
@@ -1313,6 +1313,9 @@ public class SaaSBoostInstall {
         try {
             GetParameterResponse response = ssm.getParameter(request -> request.name("/saas-boost/" + this.envName + "/SAAS_BOOST_STACK"));
             stackName = response.parameter().value();
+        } catch (ParameterNotFoundException paramStoreError) {
+            LOGGER.warn("Parameter /saas-boost/" + this.envName + "/SAAS_BOOST_STACK not found setting to default 'sb-" + this.envName + "'");
+            stackName = "sb-" + this.envName;
         } catch (SdkServiceException ssmError) {
             LOGGER.error("ssm:GetParameter error {}", ssmError.getMessage());
             LOGGER.error(getFullStackTrace(ssmError));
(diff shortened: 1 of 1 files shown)


installer/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/SaaSBoostInstall.java
```diff
@@ -1313,6 +1313,9 @@ protected String getExistingSaaSBoostStackName() {
         try {
             GetParameterResponse response = ssm.getParameter(request -> request.name("/saas-boost/" + this.envName + "/SAAS_BOOST_STACK"));
             stackName = response.parameter().value();
+        } catch (ParameterNotFoundException paramStoreError) {
+            LOGGER.warn("Parameter /saas-boost/" + this.envName + "/SAAS_BOOST_STACK not found setting to default 'sb-" + this.envName + "'");
+            stackName = "sb-" + this.envName;
```
PoeppingT, line 1318: It would be good to start maintaining all the setup defaults we have sprinkled throughout the installer in a single Defaults class or Enum to avoid potential future regressions.

installer/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/SaaSBoostInstall.java
```diff
@@ -1313,6 +1313,9 @@ protected String getExistingSaaSBoostStackName() {
         try {
             GetParameterResponse response = ssm.getParameter(request -> request.name("/saas-boost/" + this.envName + "/SAAS_BOOST_STACK"));
```
PoeppingT, line 1314: Out of scope for this PR, but these `getExisting*FromSsm()` functions could be just compiled into one `getPropertyFromSsm(String ssmPropertyName, Object default, Class returnType)` to save some code duplication.

PoeppingT (commented):
Two small non-blocking changes. If you want to separate some of this refactoring out of this PR let's create "Beginner issues" for it.

installer/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/SaaSBoostInstall.java
brtrvn: Arguably this should match the values in the SettingsService REQUIRED_PARAMS and READ_WRITE_PARAMS collections, but as the notes in the installer say we have some clean up to do on who "owns" some of these parameters (the installer, CloudFormation, the Settings service). Definitely some chicken and egg challenges. At least this path of code in the installer should only execute on an existing install.

installer/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/SaaSBoostInstall.java
brtrvn: SSM Parameter values are always String so we don't need a class argument.

PoeppingT (approved):

Issue #129, opened 2021-09-24
install_using_cloud9.md instructions installs a version of Maven incompatible with spotbugs


PoeppingT:
https://github.com/awslabs/aws-saas-boost/blob/main/docs/install_using_cloud9.md

Following the instructions in the above README installs Maven from the `amzn2-core/x86_64` yum repository, which is only version 3.0.5. SpotBugs requires Maven 3.2.5 or later.

In the documentation, 

5.  Install Apache Maven
```
sudo yum install -y maven
```

Should be replaced with

5. Install Apache Maven
```
sudo wget http://repos.fedorapeople.org/repos/dchen/apache-maven/epel-apache-maven.repo -O /etc/yum.repos.d/epel-apache-maven.repo
sudo sed -i s/\$releasever/6/g /etc/yum.repos.d/epel-apache-maven.repo
sudo yum install -y apache-maven
```

This installs Maven 3.5.2, which is compatible with SpotBugs. The above code comes from the [AWS Cloud9 Java sample for Maven](https://docs.aws.amazon.com/cloud9/latest/user-guide/sample-java.html#sample-java-sdk-maven).

---

This is a 📕 documentation issue

PR #133, opened 2021-09-27
Doc update to install correct mvn using cloud9

PoeppingT:
Fixes https://github.com/awslabs/aws-saas-boost/issues/129

----


brtrvn (commented):
Technically, Amazon Linux 2 is based on CentOS 7, so we should replace $releaseserver with 7 not 6.

Issue #130, opened 2021-09-27
Unable to install due to Uitls.jar missing in /layer/apigw-helper


MirandaDora:
The issue happens to both locally and cloud9. Unable to finish installation of SaaS boost. it fails after upload lambda functions to S3.

### Reproduction Steps

./install.sh

### What did you expect to happen?

Install SaaS Boost.

### What actually happened?

Failed installation: 
`[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 0.820 s
[INFO] Finished at: 2021-09-27T03:25:57Z
[INFO] Final Memory: 8M/30M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal on project ApiGatewayHelper: Could not resolve dependencies for project com.amazon.aws.partners.saasfactory.saasboost:ApiGatewayHelper:jar:1.0.0: Failure to find com.amazon.aws.partners.saasfactory.saasboost:Utils:jar:1.0.0 in https://repo.maven.apache.org/maven2 was cached in the local repository, resolution will not be reattempted until the update interval of central has elapsed or updates are forced -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal on project ApiGatewayHelper: Could not resolve dependencies for project com.amazon.aws.partners.saasfactory.saasboost:ApiGatewayHelper:jar:1.0.0: Failure to find com.amazon.aws.partners.saasfactory.saasboost:Utils:jar:1.0.0 in https://repo.maven.apache.org/maven2 was cached in the local repository, resolution will not be reattempted until the update interval of central has elapsed or updates are forced
    at org.apache.maven.lifecycle.internal.LifecycleDependencyResolver.getDependencies (LifecycleDependencyResolver.java:249)
    at org.apache.maven.lifecycle.internal.LifecycleDependencyResolver.resolveProjectDependencies (LifecycleDependencyResolver.java:145)
    at org.apache.maven.lifecycle.internal.MojoExecutor.ensureDependenciesAreResolved (MojoExecutor.java:246)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:200)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:154)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:146)
[...]

PR #134, opened 2021-09-28
Make installer play nice with new parent POM. Fixes Issue #130.

brtrvn:
- Update the layers modules pom to default to `clean install` instead of `clean package` so our shared libraries end up in the local maven cache.
- Update the installer to play nice with the new Boost parent pom.
- Remove old maven plugin from the ApiGatewayHelper pom.
- Cosmetic tweaks to lower the checkstyle count.
- Delete existing compiled react files from the web bucket before uploading a fresh build so you see your changes

----

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -67,7 +67,7 @@ fi
 
 cd ${CURRENT_DIR}/installer
 echo "Building installer..."
-mvn --quiet -Dspotbugs.skip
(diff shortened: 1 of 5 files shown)


PoeppingT:
Overall LGTM. Any argument with just changing the entire project's default goal to be clean install, rather than just `layers`? I can do that in the next PR I open up.

PoeppingT (approved):

Issue #144, opened 2021-10-31
SaaS Boost installation fails into new clean AWS env on SaaS Boost Metrics and Analytics CFT


LiamRandall:
The SaaS boost installation fails when trying to setup AWS SaaS Boost Metrics and Analytics Module module with an error about the CFT missing ParameterKey PrivateSubnet1 

### Reproduction Steps

New AWS Environment, Quicksight configured per guide linked here:
https://github.com/awslabs/aws-saas-boost/blob/main/docs/getting-started.md

Quicksight configured per Step-3, part 8 Standard Installation

### What did you expect to happen?

The initial installation to complete.

### What actually happened?

```
2021-10-31 10:29:30.481  INFO  SaaSBoostInstall - ===========================================================
2021-10-31 10:29:30.481  INFO  SaaSBoostInstall - Installing AWS SaaS Boost Metrics and Analytics Module
2021-10-31 10:29:30.481  INFO  SaaSBoostInstall - ===========================================================
2021-10-31 10:29:30.589  INFO  SaaSBoostInstall - Generating new random RedShift password for Analytics
2021-10-31 10:29:30.589  INFO  SaaSBoostInstall - Saving RedShift password secret to Parameter Store
2021-10-31 10:29:30.686  INFO  SaaSBoostInstall - Redshift Database User Password stored in secure SSM Parameter: /saas-boost/test/REDSHIFT_MASTER_PASSWORD
2021-10-31 10:29:30.686  INFO  SaaSBoostInstall - Creating CloudFormation stack sb-test-analytics for Analytics Module
2021-10-31 10:29:30.686  INFO  SaaSBoostInstall - Creating CloudFormation stack sb-test-analytics with database name sb_analytics_test
2021-10-31 10:29:30.686  INFO  SaaSBoostInstall - createMetricsStack::stack sb-test-analytics
2021-10-31 10:29:30.752  ERROR SaaSBoostInstall - cloudformation error
software.amazon.awssdk.services.cloudformation.model.CloudFormationException: ParameterValue for ParameterKey PrivateSubnet1 is required (Service: CloudFormation, Status Code: 400, Request ID: cf9c538b-76ad-49f3-a502-3b2a308ffab0, Extended Request ID: null)
        at software.amazon.awssdk.core.internal.http.CombinedResponseHandler.handleErrorResponse(CombinedResponseHandler.java:123) ~[SaaSBoostInstall-1.0.0-shaded.jar:?]
[...]

PR #150, opened 2021-11-02
Load base stack details if creating metrics and analytics stack during install

brtrvn:
The metrics and analytics stack is currently not a nested stack for SaaS Boost like the other modules are. The metrics and analytics stack requires parameter values from the result of the main SaaS Boost stack. When we run the installer for an update or other non-install purpose, we load the existing SaaS Boost environment data. We do not do this for the install action. If we are installing the metrics and analytics stack as part of a clean/initial installation of SaaS Boost, we need to pull the main stack details from CloudFormation when it completes prior to running the metrics and analytics stack.

Resolves issue #144 

----


PoeppingT (approved):
LGTM, confirmed this fixes the issue. Thanks for the contribution!

PR #228, opened 2022-05-27
Add Tier UI and implement default tiers.

PoeppingT:
This PR adds a new page to the Admin UI allowing SaaS Boost admins the ability to easily create, modify, and delete tiers. 

This PR also adds a "default tier" functionality as a convenience to users when adding tiers. A newly created tier will initialize with the saved AppConfig configurations of the default tier, if those configurations exist.

----

### DIFF
--- a/services/tier-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/TierService.java
+++ b/services/tier-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/TierService.java
@@ -160,6 +160,7 @@ public class TierService implements RequestHandler<Map<String, Object>, APIGatew
         Tier updatedTier = providedTier;
         try {
             Tier oldTier = store.getTier(providedTier.getId());
+            // TODO validate that user isn't trying to update fields that should not be updated, e.g. created, id
             updatedTier = store.updateTier(providedTier);
             // handling default cases:
             //   - we are now default
--- a/services/tier-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/TierService.java
+++ b/services/tier-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/TierService.java
@@ -33,21 +33,21 @@ import java.util.stream.Collectors;
 
 public class TierService implements RequestHandler<Map<String, Object>, APIGatewayProxyResponseEvent> {
     private static final Logger LOGGER = LoggerFactory.getLogger(TierService.class);
-    private static final String TABLE_NAME = System.getenv("TABLE_NAME");
+    private static final String TIERS_TABLE = System.getenv("TIERS_TABLE");
     private static final Map<String, String> CORS = Map.of("Access-Control-Allow-Origin", "*");
 
     private final TierDataStore store;
 
     public TierService() {
         final long startTimeMillis = System.currentTimeMillis();
-        if (Utils.isEmpty(TABLE_NAME)) {
+        if (Utils.isEmpty(TIERS_TABLE)) {
             throw new IllegalStateException("Missing environment variable TIERS_TABLE");
         }
         LOGGER.info("Version Info: {}", Utils.version(this.getClass()));
 
         this.store = new DynamoTierDataStore(
                 Utils.sdkClient(DynamoDbClient.builder(), DynamoDbClient.SERVICE_NAME),
-                TABLE_NAME);
+                TIERS_TABLE);
 
         LOGGER.info("Constructor init: {}", System.currentTimeMillis() - startTimeMillis);
     }
@@ -142,8 +142,8 @@ public class TierService implements RequestHandler<Map<String, Object>, APIGatew
 
         final long startTimeMillis = System.currentTimeMillis();
         Utils.logRequestEvent(event);
-        Tier updatedTier = Utils.fromJson((String) event.get("body"), Tier.class);
-        if (updatedTier == null) {
+        Tier providedTier = Utils.fromJson((String) event.get("body"), Tier.class);
+        if (providedTier == null) {
             // Utils.fromJson swallows and logs any exceptions coming from deserialization attempts
             return new APIGatewayProxyResponseEvent()
                     .withHeaders(CORS)
@@ -151,14 +151,16 @@ public class TierService implements RequestHandler<Map<String, Object>, APIGatew
                     .withBody("{\"message\":\"Body should represent a Tier.\"}");
         }
         Map<String, String> pathParams = (Map<String, String>) event.get("pathParameters");
-        if (!pathParams.get("id").equals(updatedTier.getId())) {
+        if (!pathParams.get("id").equals(providedTier.getId())) {
             return new APIGatewayProxyResponseEvent()
                     .withHeaders(CORS)
                     .withStatusCode(400)
                     .withBody("{\"message\":\"Tier IDs are immutable: body Tier ID must match path parameter.\"}");
         }
+        Tier updatedTier = providedTier;
         try {
-            Tier oldTier = store.updateTier(updatedTier);
+            Tier oldTier = store.getTier(providedTier.getId());
+            updatedTier = store.updateTier(providedTier);
             // handling default cases:
             //   - we are now default
             //     | because we enforce only one default, we must unset all other default tiers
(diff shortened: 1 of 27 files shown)


services/tier-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/model/Tier.java
```diff
@@ -39,7 +61,8 @@ public boolean equals(Object other) {
         Tier otherTier = (Tier)other;
         return this.getId().equals(otherTier.getId())
                 && this.getName().equals(otherTier.getName())
-                && this.getDescription().equals(otherTier.getDescription());
+                && this.getDescription().equals(otherTier.getDescription())
+                && this.defaultTier().equals(otherTier.defaultTier());
     }
 
     @Override
```
brtrvn, line 68: Need to add `defaultTier` to `hashCode`

services/tier-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/model/Tier.java
```diff
@@ -83,6 +108,11 @@ public Builder description(String description) {
             return this;
         }
 
+        public Builder defaultTier(Boolean defaultTier) {
+            this.defaultTier = defaultTier;
+            return this;
+        }
+
         public Tier build() {
```
brtrvn, line 116: The data access layer should be responsible for creating/maintaining the unique identifier for a persisted object not the domain model. We should implement this as we do in the rest of the SaaS Boost services were `id` is immutable and is owned by the persistence mechanism.

services/tier-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/dal/exception/TierNotFoundException.java
```diff
@@ -1,3 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package com.amazon.aws.partners.saasfactory.saasboost.dal.exception;
 
 public class TierNotFoundException extends Exception {
```
brtrvn, line 19: Prefer for this to be a RuntimeException not a checked exception

services/tier-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/dal/ddb/DynamoTierDataStore.java
```diff
@@ -74,7 +90,7 @@ public void deleteTier(String id) {
     }
 
     @Override
-    public void updateTier(Tier tier) throws TierNotFoundException {
+    public Tier updateTier(Tier tier) throws TierNotFoundException {
```
brtrvn, line 93: We don't use checked exceptions anywhere else in SaaS Boost. I'd prefer to make TierNotFoundException a runtime exception. The service code can still catch it and manage it as it wants, but let's make this a runtime exception.

services/tier-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/model/Tier.java
```diff
@@ -12,11 +28,13 @@ public final class Tier {
     private final String id;
```
brtrvn, line 28: We should add `created` and `modified` like we have with the other domain model objects in SaaS Boost to help with audit and debug. See the Tenant or User service for examples.

services/tier-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/dal/TierDataStore.java
```diff
@@ -25,7 +41,8 @@ public interface TierDataStore {
      * Updates the {@link Tier} with the same id as the provided Tier.
      *
      * @param newTier the Tier to update and the new data to supplant the old data
+     * @returns the Tier as it was before being updated
```
brtrvn, line 44: Every other service in SaaS Boost returns the updated object not the old object

services/tier-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/TierService.java
```diff
@@ -29,60 +29,36 @@
...
-     *
-     * getTiers, getTier, createTier, updateTier, deleteTier
-     * are all kind of ONLY DDB datastore actions
-     * except maybe update.
-     * either way, we should have a Tier object. JSON serializable. and DDB serializable.
-     */
     public TierService() {
         final long startTimeMillis = System.currentTimeMillis();
         if (Utils.isEmpty(TABLE_NAME)) {
             throw new IllegalStateException("Missing environment variable TIERS_TABLE");
```
brtrvn, line 44: TIERS_TABLE or TABLE_NAME? Prefer `TIERS_TABLE` to match the way we do it in the other services

services/tier-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/dal/ddb/TierAttribute.java
```diff
@@ -15,7 +31,10 @@ public enum TierAttribute {
             (tierBuilder, attributeValue) -> tierBuilder.name(attributeValue.s())),
     description(tier -> AttributeValue.builder().s(tier.getDescription()).build(),
             attributeValue -> !Utils.isEmpty(attributeValue.s()),
-            (tierBuilder, attributeValue) -> tierBuilder.description(attributeValue.s()));
+            (tierBuilder, attributeValue) -> tierBuilder.description(attributeValue.s())),
+    defaultTier(tier -> AttributeValue.builder().bool(tier.defaultTier()).build(),
```
brtrvn, line 35: Sort of a style thing, but comes from usability concerns. In all the other services, we don't use camelCase field names in the database. Instead we convert to snake_case to avoid confusion and case sensitivity issues. We don't have to change this, but for consistency we should consider it. In the future we're also going to end up with a billingPlanId -- billing_plan_id in the tiers table too.

brtrvn (commented):
Consider `Collections.singletonMap` for the `primaryKey` methods in `DynamoTier` class

PoeppingT:
@brtrvn updated the PR according to your comments, thanks for taking a look.

services/tier-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/dal/ddb/DynamoTier.java
```diff
@@ -77,11 +78,11 @@ public Map<String, AttributeValue> updateAttributes() {
...
-        return Map.of(PRIMARY_KEY.name(), attributes.get(PRIMARY_KEY.name()));
+        return Collections.singletonMap(PRIMARY_KEY.name(), attributes.get(PRIMARY_KEY.name()));
     }
 
     public static Map<String, AttributeValue> primaryKey(String id) {
-        return Map.of(PRIMARY_KEY.name(), AttributeValue.builder().s(id).build());
+        return Collections.singletonMap(PRIMARY_KEY.name(), AttributeValue.builder().s(id).build());
     }
 
     public static Tier fromAttributes(Map<String, AttributeValue> attributes) {
```
brtrvn, line 88: Something to think about... In the other services, we wrap calls to the domain object setters defensively with a check to make sure the record coming back from DDB has a value for the field before we try to deserialize it. Granted, if the current data access layer created the serialized version, the current data access layer can deserialize it without error. But, when you update and have existing records in DDB you can get NPE and other exceptions.

For example:
```
if (item.containsKey("modified")) {
    try {
        LocalDateTime modified = LocalDateTime.parse(item.get("modified").s(), DateTimeFormatter.ISO_DATE_TIME);
        tenant.setModified(modified);
    } catch (DateTimeParseException e) {
        LOGGER.error("Failed to parse created date from database: " + item.get("modified").s());
        LOGGER.error(Utils.getFullStackTrace(e));
    }
}
```

brtrvn (commented):
I think we're close. Couple more items:

The HTML rendering of the tiers page has too narrow of columns for the created and modified times. Can we make it render more like the table view of the Onboarding screen?

Need up change the IntegrationResponse method.response.header.Access-Control-Allow-Methods list for the CORS resources for /tiers and /tiers/{id}. You should list the verbs that we actually are using under that resource (OPTIONS,GET,POST in the first case and OPTIONS,GET,PUT,DELETE in the second).

Right now the UI can't update or delete because of CORS errors.

brtrvn (approved):
Looking good! Thank you for all the changes.

PR #229, opened 2022-05-31
Enabling SQS encryption at rest using the default KMS key

brtrvn:
SQS encrypts all data in transit by default. We can also encrypt data at rest on the SQS hosts while waiting to be pulled.

----


PoeppingT (approved):
LGTM, thanks Michael!

PR #230, opened 2022-06-01
Fill SSL Certificate options from ACM in Admin UI AppConfig.

PoeppingT:
This commit switches the SSL Certificate option on the AppConfig page in the Admin UI from a pure text box expecting an ACM ARN to a dropdown that is automatically filled from ACM by the Settings Service. This dropdown displays the main domain configured with the ACM certificate as well as the ARN ID in case there are multiple certificates configured with the same domain name. Currently only PENDING_VALIDATION or ISSUED certificates are displayed.

----

### DIFF
--- a/client/web/src/settings/AppSettingsSubform.js
+++ b/client/web/src/settings/AppSettingsSubform.js
@@ -55,11 +55,11 @@ export default class AppSettingsSubform extends React.Component {
                       type="select"
                       name="sslCertificate"
                       id="sslCertificate"
-                      label="ACM SSL Certificate"
+                      label={(<a href={this.props.acmConsoleLink} target="new">SSL Certificate</a>)}
                       key="sslCertificate"
                       disabled={this.props.isLocked || this.props.options?.length == 0}
                     >
-                      <option value="">{this.props.certOptions?.length == 0 ? "No Certificates in ACM!" : "Select One..."}</option>
+                      <option value="">{this.props.certOptions?.length == 0 ? "No Valid Certificates" : "Select One..."}</option>
                       {this.props.certOptions?.map((option) => (
                         <option
                           value={option.certificateArn}
--- a/client/web/src/settings/AppSettingsSubform.js
+++ b/client/web/src/settings/AppSettingsSubform.js
@@ -17,9 +17,15 @@ import { PropTypes } from 'prop-types'
 import React, { Fragment } from 'react'
 import { Row, Col, Card, CardBody, CardHeader } from 'reactstrap'
 
-import { SaasBoostInput } from '../components/FormComponents'
+import { SaasBoostInput, SaasBoostSelect } from '../components/FormComponents'
 
 export default class AppSettingsSubform extends React.Component {
+
+  certificateIdFromArn(arn) {
+    let arnParts = arn.split('/')
+    return arnParts[arnParts.length - 1]
+  }
+
   render() {
     return (
       <Fragment>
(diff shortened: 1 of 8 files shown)


services/settings-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/SettingsServiceDAL.java
```diff
@@ -164,6 +172,20 @@ public List<Map<String, Object>> rdsOptions() {
...
     }
 
+    public List<CertificateSummary> acmCertificateOptions() {
+        try {
+            // only list certificates that aren't expired, invalid, revoked, or otherwise unusable
+            ListCertificatesResponse response = acm.listCertificates(ListCertificatesRequest.builder()
+                    .certificateStatuses(List.of(CertificateStatus.PENDING_VALIDATION, CertificateStatus.ISSUED))
+                    .build());
+            LOGGER.info("ACM PENDING_VALIDATION and ISSUED certs: {}", response);
+            return response.certificateSummaryList();
```
brtrvn, line 182: We return `List.of()` in our catch, but we don't check for `response.hasCertificateSummaryList()` in the try so we theoretically could return a `null`

services/settings-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/SettingsServiceDAL.java
```diff
@@ -164,6 +172,20 @@ public List<Map<String, Object>> rdsOptions() {
         return orderableOptionsByRegion;
     }
 
+    public List<CertificateSummary> acmCertificateOptions() {
```
brtrvn, line 175: I'm assuming by leaving out `maxItems` on the request it will default to 10,000? Probably don't need to implement a while or do/while loop here with `nextToken` but I will point it out for completeness.

services/settings-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/SettingsServiceDAL.java
PoeppingT: Good callout, but it turns out that [`ListCertificatesResponse#hasCertificateSummaryList`](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/acm/model/ListCertificatesResponse.html#hasCertificateSummaryList--) "is useful because the SDK will never return a null collection or map, but you may need to differentiate between the service returning nothing (or null) and the service returning an empty collection or map".

In this case we don't need to differentiate: the service responding with no certificate summary list (but also no exception) is functionally the same as returning an empty list. Plus, documentation says the SDK will never return a null collection/map.

But, I agree that we should `null` protect, just in case. So I'll implement that in the next diff.

services/settings-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/SettingsServiceDAL.java
PoeppingT: You're right, correctness requires we loop on `nextToken`. Good catch!

client/web/src/settings/AppSettingsSubform.js
```diff
@@ -45,13 +51,24 @@ export default class AppSettingsSubform extends React.Component {
...
-                      key="sslCertificate"
-                      label="SSL Certificate ARN"
+                    <SaasBoostSelect
+                      type="select"
                       name="sslCertificate"
-                      type="text"
-                      disabled={this.props.isLocked}
-                    />
+                      id="sslCertificate"
+                      label="ACM SSL Certificate"
```
brtrvn, line 58: Could be more user friendly to have this label be "SSL Certificate" and have it be a hyperlink to the Certificate Manager console https://us-west-2.console.aws.amazon.com/acm/home?region=us-west-2#/certificates/list (obviously will have to dynamically set the region in that URL)

brtrvn (approved):
Looks good

PR #239, opened 2022-06-27
Move database configuration to services, performance characteristics (instanceClass) to tier.

PoeppingT:
This commit refactors the way that AppConfig is saved for databases:
moving the database configuration object to be service specific and
allowing only the database performance characteristics to be
configurable by tier, replacing the existing logic where all database
configurations were configurable by tier. Note that this brings with it
the restriction that now services cannot be created where a difference
of tier leads to a difference of database version, engine, or existence.

This commit also reworks the UI to support the above changes and the
RdsOptions custom resource to more accurately represent RDS built-in
dependencies when selecting database configurations, as well as changing
the Onboarding service to correctly select from the AppConfig object.

----

### DIFF
--- a/client/web/src/settings/ApplicationComponent.js
+++ b/client/web/src/settings/ApplicationComponent.js
@@ -73,8 +73,8 @@ export function ApplicationComponent(props) {
   const generateAppConfigOrDefaultInitialValuesForTier = (tierValues, defaultValues, fileSystemType) => {
     let tierValuesCopy = Object.assign({}, tierValues)
     let defaults = Object.assign({
-      min: 1,
-      max: 1,
+      min: 0,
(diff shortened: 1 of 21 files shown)


brtrvn (commented):
Maybe we should change the wording on the UI to say "Provision a database for this service" instead of "...for the application".

In the onChange or onBlur or whatever it is we're capturing when the tier combo box is changed, the database instance type is copied over from default, but the compute size and instance counts are not.

Since we're in here, we should change the parameter name for the password to remove the non-inclusive "master".

brtrvn (approved):
Looks good. Let's capture the SQL bootstrap file upload issue so we can dig into it later.

Issue #231, opened 2022-06-03
error shows up in cloudwatch when onboard tenants: Can't find onboarding record for tenant tenant


qwudssd:
Tenants could not be successfully deployed and there is an error in onboarding-events in CloudWatch:

<img width="1321" alt="Screen Shot 2022-06-03 at 11 47 18 AM" src="https://user-images.githubusercontent.com/104920467/171900481-afab1740-4262-491f-b97b-b48652e0e36b.png">

### Environment

  - **AWS SaaS Boost Version:** Saas main branch

qwudssd:
thanks for the response! hope the info below helps
1. environment name: dev-t15
    Region: us-west-2
2. tenant’s name: test
    Used the provision tenant feature in the user interface to provision the tenant
3. Haven’t made any changes in the local SaaS Boost
4. the entire log that has the error msg:
<img width="1351" alt="Screen Shot 2022-06-03 at 2 56 20 PM" src="https://user-images.githubusercontent.com/104920467/171941533-0d2a9f52-040e-493a-8245-595146d76652.png">

5. the event that was being handled is: OnboardingService - Handling Onboarding Deployment Pipeline Changed

PoeppingT:
This is happening due to the `-` character in your environment name. To unblock yourself immediately, create a new environment without that character to continue testing and we will address this bug when we can.

qwudssd:
okay i see. Thanks!

rnlduaeo:
Please review the PR. https://github.com/awslabs/aws-saas-boost/pull/240

rnlduaeo:
@Poepping When you change this bug, please make sure that env name is also used in athena query table, when I manually changed the split function in /aws-saas-boost/services/onboarding-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/OnboardingService.java line 1111 > tenantId = pipeline.split("-")[3]; 
to tenantId = pipeline.split("-tenant-")[1].split("-")[0]; 
The tenant onboarding works, but the request count metric values were not fetched in the 'Usage' page. This was due to the fact that env name is also used in athena query table, I thought about wrapping it in double quotation but was not sure about any downstreaming effect of changing that. Please consider this as well when you fix the bug.

PR #240, opened 2022-06-29
Modify environment name validation check

rnlduaeo:
----

Existing environment name check condition allows special characters, however environment name is being used in athena table and other string split (by '-') functions. I modified validation check method and corresponding output messages. That's all.

Below issue has been published. This PR will prevent the user from typing special characters for environment name.
https://github.com/awslabs/aws-saas-boost/issues/231


qwudssd (approved):

brtrvn:
The downstream effect of the dash character in the environment label is the overly simplistic logic we are using to identify the CodePipeline for a tenant based on meta data available in the eventing system. The proper fix for this isn't to exclude the dash character from the environment label, but to fix the logic that figures out the proper CodePipeline for a tenant. This work is in progress already.

rnlduaeo:
From what I found so far, not only identifying codepipeline, but also identifying athena table name (dash is not allowed in table name). So if your team decided to change the logic, then other downstream effects should be considered and fixed.

PR #248, opened 2022-07-08
Simplify the filesystem JSON in AppConfig.

PoeppingT:
This commit refactors the SharedFileSystem class out of the Settings
Service, replacing it with new FileSystems classes that inherit relevant
fields for CloudFormation templates from each other. For example, all
FileSystems have a mount point, but only FSX filesystems support
specifying the amount of storage you want. This also flattens the
filesystem objects in the AppConfig, getting rid of individual "efs" and
"fsx" objects and replacing them instead with just the existing
"filesystem" object, including or omitting fields as necessary.

This commit also completes a similar refactoring in the AppConfig UI,
making it significantly easier to add FileSystems in the future, as well
as fixing various small bugs in the AppConfig UI.

----


brtrvn (approved):
Looks like a good pattern

PR #249, opened 2022-07-16
Fix bug selecting tier for provision tenant and creating a new defaul…

PoeppingT:
…t tier.

----


brtrvn (approved):
Looks good. As discussed, _writing_ to the tier service should be a low TPS/low concurrency action (in fact it may only ever happen once), so the overhead of introducing DynamoDB transactions is not a good investment right now.

Issue #241, opened 2022-06-30
InvokeRDSBootstrapDatabase throws the error while trying to connect to database


rnlduaeo:
Database initialization functionality in SaaSBoost tried to access poastgresql and threw an exception. I got below error message in rdsbootstrap lambda log group.
<img width="1161" alt="image" src="https://user-images.githubusercontent.com/5877925/176625587-81eaf65b-7581-4f20-b902-dff4b55a1732.png">

### Reproduction Steps

1) upload .sql file(UTF-8 encoded) in rds appconfig screen.
2) try to onboard a tenant, you will see the above error message.

### What did you expect to happen?

At least the database connection should be established, even though some syntax error might be occurred.

### What actually happened?

Can't connect to database host

### Environment

  - **AWS Region  :** ap-northeast-2
  - **AWS SaaS Boost Version :** Version 798371c
  - **Workload OS  (Linux or Windows) :**  linux

### Other

--- 

This is :bug: Bug Report

bella-kwon:
For someone who encounters this error like me.
Here's the root cause I found -
ResultSet had been closed within try block(line: 187) in [RdsBootstrap.databaseExists](https://github.com/awslabs/aws-saas-boost/blob/798371c40b032e097d51ac9e57292af78bc201c4/resources/custom-resources/rds-bootstrap/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/RdsBootstrap.java#L181) before trying rs.next() (line:194)
I So I just moved while block while (rs.next()) {  into the try block as follows. You can fix this temporarily by changing this part and try to run it.
![image](https://user-images.githubusercontent.com/43110337/176792754-4abcc5cf-0b8d-4c19-9004-e6bf2b0d62af.png)

Of course, I understand that there will be many considerations for supporting DB engines from the team's perspectives.

I wrote this to reduce the # of developers who are wandering until then.

Issue #250, opened 2022-07-16
RDS bootstrap code fails with SQL files including semicolons not at the end of a statement


brtrvn:
The RDS bootstrap code parses the SQL text file into statements to send to the database based on semicolons. The code does not parse the file line-by-line, because SQL statements often span multiple lines. Because we do not parse the file line-by-line, we do not ignore SQL comment lines. The comments are sent to the database backend via the JDBC driver along with the rest of the SQL statement.

If there is a semicolon that does not designate the end of a SQL statement to be sent to the database, the bootstrap will fail. By default, the PostgreSQL backup utility program `pg_dump` creates comments to describe the database objects it is writing to file. These comments include semicolon characters and break the RDS bootstrap parser.

See Issue #246 

### Reproduction Steps
Setup a Postgres database for an application service and upload a UTF-8 encoded SQL file with a semicolon in the middle of a line. For example:
```sql
--
-- Name: foo; Type: TABLE; Schema: public; Owner: postgres
--

CREATE TABLE public.foo (
    bar character varying(255) NOT NULL
);
```
### What did you expect to happen?
SaaS Boost should bootstrap my service's database with a table called `foo` that has a single VARCHAR column `bar`

### What actually happened?
Tenant onboarding fails because the RDS bootstrap custom resource Lambda fails.

### Environment

  - **AWS Region  :**
  - **AWS SaaS Boost Version :**
  - **Workload OS  (Linux or Windows) :**

### Other

--- 

This is :bug: Bug Report

PR #251, opened 2022-07-16
RDS Bootstrap fixes

brtrvn:
Fixes issues #241 and #250.

Update the RDS bootstrap file scanner to parse on semicolon characters only when they are at the end of a line.
Update the database existence check code to not use try with resources which was closing out the `ResultSet` before we could read it when connecting to a Postgres database.

----


PoeppingT (approved):
LGTM

PR #259, opened 2022-07-26
Allow for EC2 Launch Type with Linux-based services.

PoeppingT:
This commit also fixes a small unreported bug where the onboarding
service is unable to delete the LinuxLaunchTemplate by adding that
permission to the onboarding service role.

----

### DIFF
--- a/client/web/src/settings/ServiceSettingsSubform.js
+++ b/client/web/src/settings/ServiceSettingsSubform.js
@@ -58,7 +58,7 @@ const ServiceSettingsSubform = (props) => {
   const getLaunchTypeOptions = (serviceIndex) => {
     return serviceValues?.operatingSystem === 'LINUX' && (
      <FormGroup>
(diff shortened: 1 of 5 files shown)


brtrvn (commented):
The label above the launch type radio buttons in the UI should say "Container Launch Type"

brtrvn:
Looks like a small bug in the changes to Onboarding for the FSx file system types. Obviously provisioning a filesystem is optional, but the template has validation to only allow `FSX_WINDOWS` or `FSX_ONTAP` as values for the `FSxFileSystemType` parameter. We should default the `fileSystemType` variable to "FSX_WINDOWS" instead of a blank string to match the template default. Right now, if you don't provision a file system with your service, it will error out.

brtrvn (approved):
Looks good. We will address the default file system type issue in a different PR associated with FSx for ONTAP support.

PR #261, opened 2022-07-28
Support FSX Ontap in AppConfig.

PoeppingT:
This commit also fixes a bug where an application configuration created
with no filesystem will fail to onboard tenants.

Co-authored-by: netapp-vedantsethia <Vedant.Sethia@netapp.com>
Co-authored-by: netapp-dhruv-tyagi <Dhruv.Tyagi@netapp.com>

----


brtrvn (commented):
Couple of inconsistencies between equals and hashcode and we need to make sure we're setting the WIndows mount drive letter for both FSx Windows and FSx ONTAP.
I'll push small updates to address those and then we can merge.

brtrvn (approved):

PR #263, opened 2022-07-29
Add UI support for selecting FSX Ontap for Windows Applications.

PoeppingT:
This commit also fixes a bug with the update_local_env.sh utility
script.

Co-authored-by: netapp-vedantsethia <Vedant.Sethia@netapp.com>
Co-authored-by: netapp-dhruv-tyagi <Dhruv.Tyagi@netapp.com>

----

Screen Shot 2022-07-29 at 11 14 35

### DIFF
--- a/client/web/src/settings/ApplicationComponent.js
+++ b/client/web/src/settings/ApplicationComponent.js
@@ -73,7 +73,7 @@ export function ApplicationComponent(props) {
     let filesystem = {
       ...FILESYSTEM_DEFAULTS,
       ...defaultValues.filesystem,
-      ...tierValues.filesystem
+      ...splitWeeklyMaintenanceTime(tierValues.filesystem)
     }
     let defaults = Object.assign({
       min: 0,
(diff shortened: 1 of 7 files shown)


brtrvn (commented):
The visibility of the radio buttons for FSx file system type looks to be controlled by the container OS radio button selection not the checkbox for whether you want to configure a file system or not.
Looks like we lost the FSx Windows defaults for throughput and storage and now show the FSx ONTAP defaults for both types.

brtrvn (commented):
Looks like `ecsLaunchType` is left NULL when you select Windows for your Container OS. We can either show the Container Launch Type radio buttons for WIndows, set it to EC2 (for now) and disable it so the user can't change it, or we can set it on the backend to EC2 instead of NULL so onboarding doesn't fail.

brtrvn:
Resolved the `ecsLaunchType` issue in #270 because it looks like it came up with the EC2 support for Linux workloads not specifically with the FSx ONTAP changes.

brtrvn (approved):
Looks good. Thank you for fixing the maintenance start time bug. And, I like that we're getting setup to support FSx ONTAP on Linux workloads launched on EC2!

PR #267, opened 2022-08-08
Replace authorizer to lambda CUSTOM authorizer

amliuyong:
- created a lambda function to do authorization
- created CUSTOM authorizer in API gateway
- replaced Cognito authorizer with CUSTOM authorizer


brtrvn (commented):
This is a great start, but I think it's overly complicated for what we need to do right now.

I think we can simplify this PR and make our move to a configurable OIDC provider more incremental and easier to test/evaluate and incorporate into the main line of SaaS Boost.

- No need to add anything to the installer right now. All we need to do is replace the built-in Amazon API Gateway Cognito User Pool authorization with a Lambda Authorizer that does the exact same thing.
- You're not using any of the Utility methods from the SaaS Boost Utils Lambda layer.
- Looks like we have a mix of the Java JWT library and the Auth0 JWT library. They are both active projects. JJWT has been around longer and is Apache 2.0. The Auth0 library is MIT licensed.
- We should simply called the Lambda authorizer "authorizer" What if we want to support non OIDC authorization schemes with our custom Lambda authorizer?
- The AuthPolicy class looks like it's come from some other author/project? What's the origin of this? 
- There are a number of commented out chunks of code in various files that we should clean up

amliuyong:
> This is a great start, but I think it's overly complicated for what we need to do right now.
> 
> I think we can simplify this PR and make our move to a configurable OIDC provider more incremental and easier to test/evaluate and incorporate into the main line of SaaS Boost.
> 
> * No need to add anything to the installer right now. All we need to do is replace the built-in Amazon API Gateway Cognito User Pool authorization with a Lambda Authorizer that does the exact same thing.
> * You're not using any of the Utility methods from the SaaS Boost Utils Lambda layer.
> * Looks like we have a mix of the Java JWT library and the Auth0 JWT library. They are both active projects. JJWT has been around longer and is Apache 2.0. The Auth0 library is MIT licensed.
> * We should simply called the Lambda authorizer "authorizer" What if we want to support non OIDC authorization schemes with our custom Lambda authorizer?
> * The AuthPolicy class looks like it's come from some other author/project? What's the origin of this?
> * There are a number of commented out chunks of code in various files that we should clean up

thank you [brtrvn](https://github.com/brtrvn).

I made some changes based on your comments.

1.  undo changes in installer.  please note that Lambda Authorizer verify jwt token based on OIDC, which is different with current Cognito User Pool authorization, it will break current application.  More changes (client/web, installer)need to be made to make it work, these changes will be submitted in other PRs. 
2.  change to use Utils Lambda layer in Lambda Authorizer
3. yes,  have a mix of the Java JWT library and the Auth0 JWT library.  Use Auth0 JWK to get public key, and use JJWT to  parse JWT token (JJWT does not have JJWT api). 
4. changed lambda name to "authorizer"
5. AuthPolicy is from https://github.com/awslabs/aws-apigateway-lambda-authorizer-blueprints, and I made a little change to it(because it does not works in China regions) . 
6. removed commented out chunks.

brtrvn:
Superseded by #368

PR #268, opened 2022-08-09
Service shutdown and startup v2

brtrvn:
Updates to the service shutdown and startup functions to support multiple services in a v2 app config.

----


brtrvn (commented):
We could be more defensive testing for unexpected states like missing tier configurations, mismatch on service names in ECS vs what's in the app config, etc... but, what would it mean to have some tenants running and others not or some services running and others not?

PR #270, opened 2022-08-09
Default ecsLaunchType to EC2 for Windows based services

### LINKED ISSUE #269: Windows based services fail to onboard with NULL ecsLaunchType
We recently added support for choosing either Fargate or EC2 launch types for Linux based workloads. The launch type selection is only displayed if you've selected Linux as your container's operating system. If you choose Windows, the launch type selection is not shown and it is not defaulted to EC2. A `null` launch type is saved with the application config and onboarding fails.

### Reproduction Steps
Configure an application service to use Windows for the operating system
Push an image to the service's ECR repo
Onboard a tenant
Onboarding will fail when provisioning the infrastructure for the service

Onboarding succeeds and the Windows based service launched on EC2 container instances by ECS (not Fargate)

### What actually happened?
Onboarding fails

### Environment

  - **AWS Region  :** Any
  - **AWS SaaS Boost Version :** HEAD
  - **Workload OS  (Linux or Windows) :** Windows

### Other

--- 

This is :bug: Bug Report

brtrvn:
If the `ecsLaunchType` for a service is not defined, set it to `FARGATE` when the container operating system is Linux and `EC2` when Windows

Resolves #269 

----

### DIFF
--- a/client/web/src/settings/ApplicationContainer.js
+++ b/client/web/src/settings/ApplicationContainer.js
@@ -209,6 +209,7 @@ export function ApplicationContainer(props) {
           name,
           windowsVersion,
           operatingSystem,
+          ecsLaunchType,
           provisionDb,
           tombstone,
           database,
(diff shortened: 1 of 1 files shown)


brtrvn (commented):
Maybe a better UX here would be to show the launch type radio buttons for Windows, but pre-select EC2 and disable the control so it's clear what's going on. When we add Fargate support for Windows then we simply enable the launch type radio buttons.

PR #271, opened 2022-08-10
Prevent CloudFormation parameter errors when using EFS

brtrvn:
The CloudFormation template parameters for FSx file systems limit the FSx type to either Windows File Server or NetApp ONTAP. When we onboard a service with EFS instead of FSx, the file system type is passed through to CloudFormation and fails the parameter required values check.

----


brtrvn (commented):
The other options would be to remove the `AllowedValues` constraint from tenant-onboarding-app.yaml but leave it in tenant-onboarding-fsx.yaml, or to add `EFS` to the allowed values in the outer template.

PR #274, opened 2022-08-11
Allow dash in SaaS Boost environment name

brtrvn:
Update the onboarding service CodePipeline listener to be more precise by fetching the tenant id from the pipeline tags instead of heuristics based on the pipeline name. Also, quote the table name (which includes the SaaS Boost environment label) in the Athena SQL queries for the Metrics service.

This resolves issue #231 when the SaaS Boost environment contains a hyphen/dash character.

----


brtrvn (commented):
Note that tagging the CodePipeline resource has been silently failing without the IAM permissions to codepipeline:TagResource

Single quoting the table name for the Athena query did not work.

PR #275, opened 2022-08-12
CloudFormation custom resource response retry

### LINKED ISSUE #273: During install the InvokeClearWebsiteBucket custom resource fails to create
I am trying to deploy a new installation but the install.sh script failed.

The CloudFormation error for the InvokeClearWebsiteBucket custom resource:

`CloudFormation did not receive a response from your Custom Resource. Please check your logs for requestId [aed836fa-a458-4b84-97de-f6b7243fe992]. If you are using the Python cfn-response module, you may need to update your Lambda function code so that CloudFormation can attach the updated version.`

### Reproduction Steps

I followed the instructions for installing from Cloud9.

The installation would complete without errors.

### What actually happened?

The installation failed.

### Environment

  - **AWS Region  : us-east-1**
  - **AWS SaaS Boost Version : Installer Version 9793e06 **
  - **Workload OS  (Linux or Windows) : NA**

### Other

--- 

This is :bug: Bug Report

brtrvn:
If the call to PUT on the CloudFormation S3 bucket for custom resources fails, retry with exponential backoff and jitter. After 5 attempts, fail the call. Resolves #273. 

----

### DIFF
--- a/layers/cloudformation-utils/pom.xml
+++ b/layers/cloudformation-utils/pom.xml
@@ -33,7 +33,7 @@ limitations under the License.
     </licenses>
 
     <properties>
-        <checkstyle.maxAllowedViolations>7</checkstyle.maxAllowedViolations>
+        <checkstyle.maxAllowedViolations>0</checkstyle.maxAllowedViolations>
     </properties>
 
     <build>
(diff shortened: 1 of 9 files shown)


brtrvn (commented):
Would be nice to confirm the fix, but the error is so intermittent. We can also probably tweak the retry function inputs so we're not waiting for any longer than we need to.

PR #276, opened 2022-08-16
Add Tier to tag setting

muylucir:
----
Cost analysis is a very important area in SaaS.
SaaS Boost supports various cost allocation tags.
However, the cost allocation tag for the tier is not supported.
Accurate cost measurement for each tier is an essential element in cost estimation.

For example, if a lower tier spends more money than a higher tier, it will adversely affect profitability. We should be able to identify this.

Screen Shot 2022-08-16 at 11 49 24 AM

In the process of creating a tenant, a code was added to add the tier information to the tag and use it for cost estimation.

It's a simple fix, but I think it's necessary.

### DIFF
--- a/resources/tenant-onboarding-app.yaml
+++ b/resources/tenant-onboarding-app.yaml
@@ -21,6 +21,10 @@ Parameters:
   TenantId:
     Description: The GUID for the tenant
     Type: String
+  Tier:
+    Description: The tier this tenant is onboading into
+    Type: String
+    Default: ''
   ServiceName:
     Description: Name for this application service
     Type: String
@@ -288,6 +292,8 @@ Resources:
       Tags:
         - Key: Tenant
           Value: !Ref TenantId
+        - Key: Tier
+          Value: !Ref Tier
   ClusterCapacityProviderAssociations:
     Type: AWS::ECS::ClusterCapacityProviderAssociations
     Condition: Ec2LaunchType
@@ -459,6 +465,8 @@ Resources:
       Tags:
         - Key: Tenant
           Value: !Ref TenantId
+        - Key: Tier
+          Value: !Ref Tier
       ContainerDefinitions:
         - Name:
             Fn::Join: ['', ['sb-', !Ref Environment, '-tenant-', !Select [0, !Split ['-', !Ref TenantId]], '-', !Ref ServiceResourceName]]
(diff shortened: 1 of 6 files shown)


PoeppingT (changes requested):
Hello @muylucir , thanks for submitting this PR! I have a few changes to suggest:
1. in `tenant-onboarding-app.yaml` you have not added the Tier tag to the CodePipeline and WindowsLaunchTemplate resources. If this was intentional, can you please explain why?
2. the Tier Parameter needs to be passed to each of the sub-stacks: `efs`, `fsx`, and `rds`. This can be done by adding `Tier: !Ref Tier` to the `Parameters` those resources at the bottom of [`tenant-onboarding-app.yaml`](https://github.com/awslabs/aws-saas-boost/blob/main/resources/tenant-onboarding-app.yaml#L1144-L1205)
3. the `tenant-onboarding-app.yaml` CloudFormation stack is created by the SaaS Boost Onboarding Service. It's a good idea to add Tier as a parameter to that stack, but for that parameter to have any function it must be passed by the Onboarding Service. Look for the [`handleOnboardingBaseProvisioned`](https://github.com/awslabs/aws-saas-boost/blob/main/services/onboarding-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/OnboardingService.java#L715) function in `services/onboarding-service/.../OnboardingService.java` to find how those stack parameters are collected and how the stack is created in response to provisioning events.

Lastly I would recommend you test to validate your changes. This can be done by following the sample application instructions in the [Getting Started Guide](https://github.com/awslabs/aws-saas-boost/blob/main/docs/getting-started.md) and validating that each resource you want to be tagged per tenant is tagged with the proper Tier. 

Thanks for your contribution, and I look forward to seeing your changes soon!

muylucir:
Hi @PoeppingT 
Thanks for the detailed explanation. 

A1) No other reason. I think I left it out by mistake.
A2) I understand how to pass parameters to other resources in tenant-onboarding-app.yaml.
A3) I'll take a look at handleOnboardingBaseProvisioned again and understand.

I will request a review after making necessary changes and testing enough.

Thanks!! :)

muylucir:
Hi @PoeppingT 

I modified the code. To pass the Tier tag as a parameter to the tenant-onboarding-app.yaml CloudFormation stack through handleOnboardingBaseProvisioned in services/onboarding-service/.../OnboardingService.java.

 I used the sample application to complete the necessary tests. Tier tags are normally assigned to all resources (rds, efs, ecs, etc.) created by CloudFormation stack.

PR #284, opened 2022-08-25
Bring Utils into installer to enable intelligent installer update

PoeppingT:
----

### DIFF
--- a/installer/pom.xml
+++ b/installer/pom.xml
@@ -109,6 +109,17 @@ limitations under the License.
 
     <dependencies>
         <dependency>
+            <groupId>com.amazon.aws.partners.saasfactory.saasboost</groupId>
+            <artifactId>Utils</artifactId>
+            <version>1.0.0</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>software.amazon.awssdk</groupId>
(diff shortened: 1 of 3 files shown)


brtrvn (changes requested):
We can replace `SaaSBoostInstall::generatePassword` with `Utils::randomString(16)`

Can use `Utils::isBlank` in `getExistingSaaSBoostEnvironment`

Optionally, we could also update the S3 artifacts bucket naming to use `randomString(12, "[^a-z0-9]")` for the suffix to match what CloudFormation would have done instead of what we're currently doing with UUID.

PoeppingT:
> We can replace `SaaSBoostInstall::generatePassword` with `Utils::randomString(16)`

Our Cognito UserPool password policy currently requires at least one lowercase letter, uppercase letter, and number. `Utils::randomString` does not guarantee this in the same way that `SaaSBoostInstall::generatePassword` does. I will alter `SaaSBoostInstall::generatePassword` to conform to that policy while still using `Utils::randomString`.

brtrvn (commented):
Looks good. In the future, we can look at adding required character type constraints to the Utils randomString method.

brtrvn (approved):
Looks good

PR #285, opened 2022-08-30
Fix installing SaaS Boost without first running a mvn build.

PoeppingT:
Recent changes to the installer brought in a dependency on the layers/utils package in SaaS Boost. Because the install scripts have always only built the installer package, this would break for those people that have never built the main parent pom, layers parent pom, and utils package itself. 

This commit adds in explicit builds into the install scripts to ensure those are built before trying to build the installer.

----


brtrvn (approved):
LGTM

PR #293, opened 2022-09-16
Improve UX on invalid AppConfig submission.

PoeppingT:
This commit changes the AppConfig page on the Admin UI with three small experience changes:
1. When clicking submit, the page will scroll to the top. This makes any errors evident in the banner at the top of the page.
2. Banners have been updated to the newer Bootstrap format, meaning the error banner is now properly 'dangerous'.
3. The error banner at the top now includes not only which service has errors, but which tiers in those services as well.

----

### DIFF
--- a/client/web/src/settings/ApplicationComponent.js
+++ b/client/web/src/settings/ApplicationComponent.js
@@ -367,16 +367,17 @@ export function ApplicationComponent(props) {
       } else {
         formik.errors.services?.forEach((service, index) => {
           if (!!service) {
-            let serviceString = formik.values?.services[index].name
+            let serviceName = formik.values?.services[index].name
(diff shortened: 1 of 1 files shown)


PoeppingT:
An example of the new UX: note the red banner and the inclusion of tiers in the error popup.

Screen Shot 2022-09-16 at 12 24 07

brtrvn (commented):
Should we explicitly label what we're showing them as tiers to make the error message more precise, or just figure it will be obvious because they named the tiers. Wondering if people will be confused as to what `default` means if they don't ever touch tier config.

brtrvn (approved):
Looks ok for now. Thanks for improving the user experience!

PR #303, opened 2022-09-26
Don't hardcode AWS partition and URL suffix

brtrvn:
Update SaaS Boost to variably refer to the AWS partition and the AWS URL suffix to support running in non US commercial regions. This includes both using the AWS::Partition and AWS::URLSuffix pseudo parameters in CloudFormation as well as updating the Utils library to properly create SDK clients with the correct service endpoints.

Also updating inline IAM policies created as part of an IAM role to drop the AWS region name since inline policies are encompassed by the containing role so their names do not need to be unique as long as the role names are unique.

Co-authored-by: amliuyong <yonmzn@amazon.com>
Co-authored-by: PoeppingT <poeppt@amazon.com>

----

### DIFF
--- a/resources/saas-boost-svc-onboarding.yaml
+++ b/resources/saas-boost-svc-onboarding.yaml
@@ -347,7 +347,7 @@ Resources:
             Action:
               - sns:Publish
             Resource:
-              - !Sub arn:${AWS::Partition}:sns:${AWS::Region}:${AWS::AccountId}:sb-*
+              - !Sub arn:${AWS::Partition}:sns:${AWS::Region}:${AWS::AccountId}:sb-${Environment}-onboarding*
           - Effect: Allow
             Action:
               - ssm:GetParameter*
--- a/resources/saas-boost-svc-onboarding.yaml
+++ b/resources/saas-boost-svc-onboarding.yaml
@@ -172,26 +172,26 @@ Resources:
             Action:
               - sts:AssumeRole
       Policies:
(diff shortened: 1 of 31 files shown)


resources/saas-boost-svc-onboarding.yaml
```diff
@@ -340,41 +340,41 @@ Resources:
...
-              - !Sub arn:aws:cloudformation:${AWS::Region}:${AWS::AccountId}:stack/sb-${Environment}*
-              - !Sub arn:aws:cloudformation:${AWS::Region}:${AWS::AccountId}:transform/saas-boost-app-services-macro
+              - !Sub arn:${AWS::Partition}:cloudformation:${AWS::Region}:${AWS::AccountId}:stack/sb-${Environment}*
+              - !Sub arn:${AWS::Partition}:cloudformation:${AWS::Region}:${AWS::AccountId}:transform/saas-boost-app-services-macro
           - Effect: Allow
             Action:
               - sns:Publish
             Resource:
-              - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:sb-*
+              - !Sub arn:${AWS::Partition}:sns:${AWS::Region}:${AWS::AccountId}:sb-*
```
PoeppingT, line 349: It looks like all the SNS topics we create are of the form `sb-${Environment}-*`. Can we change this to be more restrictive?

PoeppingT (changes requested):
Should we update the sample app build scripts as well?

PoeppingT (approved):
LGTM!

PR #305, opened 2022-09-27
Fix bug deleting SaaS Boost created Route53 Hosted Zone on AppConfig …

### LINKED ISSUE #300: Updating AppConfig deletes SaaS Boost created HostedZone
If you have a SaaS Boost environment that automatically created a HostedZone because it was configured with a DomainName, updating the AppConfig afterwards (even if just updating something like a service description) will result in the HostedZone getting deleted.

### Reproduction Steps

1. Stand up a SaaS Boost environment
2. Configure it with a Domain Name and 1 service
3. See that a HostedZone is created in Route53 with the provided Domain Name as its name and the description `"${DomainName} Public DNS Zone"`
4. Update the service description
5. See that Settings service updates the config and fires an "Application Configuration Changed" event in the `sb-${env}-settings-update-config` CloudWatch LogGroup
6. See that the Onboarding service finds the existing HostedZone and updates the base stack in `sb-${env}-onboarding-events`
7. See that the HostedZone is eventually deleted.

I expect that the HostedZone would stick around.

### What actually happened?

The HostedZone was deleted.

### Environment

  - **AWS Region  :** us-west-2
  - **AWS SaaS Boost Version :** `main` at https://github.com/awslabs/aws-saas-boost/commit/176187a405f9ce574544745ed8ec01e0db7172dd
  - **Workload OS  (Linux or Windows) :** N/A

### Other

This is because the condition on which the HostedZone is created is based on whether a HostedZone is passed (the intention here was to only create a HostedZone if one does not exist): [source](https://github.com/awslabs/aws-saas-boost/blob/main/resources/saas-boost-core.yaml#L64-L67)
```
# in resources/saas-boost-core.yaml
Conditions:
  CreateHostedZone: !And
    - !Not [!Equals [!Ref DomainName, '']]
    - !Equals [!Ref HostedZone, '']
```

However, the Onboarding Service sets the `HostedZone` parameter when updating the stack to whatever exists in Route53, which will be the Boost-created HostedZone if none already existed: [source](https://github.com/awslabs/aws-saas-boost/blob/main/services/onboarding-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/OnboardingService.java#L1771-L1801)
```
String hostedZone = getExistingHostedZone(domainName);

// If there's an existing hosted zone, we need to tell the AppConfig about it
// Otherwise, if there's a domain name, CloudFormation will create a hosted zone
// and the stack listener will tell AppConfig about the newly created one.
if (Utils.isNotBlank(hostedZone)) {
  LOGGER.info("Publishing appConfig update event for Route53 hosted zone {}", hostedZone);
  Utils.publishEvent(eventBridge, SAAS_BOOST_EVENT_BUS, EVENT_SOURCE,
    "Application Configuration Resource Changed",
    Map.of("hostedZone", hostedZone));
}

...

Parameter.builder().parameterKey("HostedZone").parameterValue(hostedZone).build(),
```

Using the `getExistingHostedZone` function: [source](https://github.com/awslabs/aws-saas-boost/blob/main/services/onboarding-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/OnboardingService.java#L2188-L2226). It's worth noting that this is also where we only reuse existing HostedZones if SaaS Boost already created them, an issue brought up by users in https://github.com/awslabs/aws-saas-boost/discussions/299 and https://github.com/awslabs/aws-saas-boost/issues/257.
```
protected String getExistingHostedZone(String domainName) {
    String existingHostedZone = "";
    if (Utils.isNotEmpty(domainName)) {
        String nextDnsName = null;
        String nextHostedZone = null;
        ListHostedZonesByNameResponse response;
        do {
            response = route53.listHostedZonesByName(ListHostedZonesByNameRequest.builder()
                    .dnsName(nextDnsName)
                    .hostedZoneId(nextHostedZone)
                    .maxItems("100")
                    .build()
            );
            nextDnsName = response.nextDNSName();
            nextHostedZone = response.nextHostedZoneId();
            if (response.hasHostedZones()) {
                for (HostedZone hostedZone : response.hostedZones()) {
                    // If there are multiple hosted zones for a given domain name, what should we do?
                    // We could sort the response by "CallerReference" which appears to be a timestamp.
                    // In the documentation, we can just tell people if they're suffering from
                    // https://github.com/awslabs/aws-saas-boost/issues/74 to go clean things up manually first?
                    if (hostedZone.name().startsWith(domainName)
                            && hostedZone.config() != null
                            && Boolean.FALSE.equals(hostedZone.config().privateZone())) {
                        // Created by SaaS Boost CloudFormation?
                        // TODO do we do this check? seems safest for now.
                        if ((domainName + " Public DNS zone").equals(hostedZone.config().comment())) {
                            LOGGER.info("Found existing hosted zone {} for domain {}", hostedZone, domainName);
                            // Hosted zone id will be prefixed with /hostedzone/
                            existingHostedZone = hostedZone.id().replace("/hostedzone/", "");
                            break;
                        }
                    }
                }
            }
        } while (response.isTruncated());
    }
    return existingHostedZone;
}
```

When the Onboarding Service updates the stack passing in the created HostedZone id, the condition evaluates to `false` and CloudFormation deletes the HostedZone.

One possible way to fix this is to view the existing stack resources for the core stack before passing template parameters. Pseudo-code follows:
```
if appConfig.domainName:
  stackParameters.domainName = appConfig.domainName
  stackResources = cfn.getStackResources(sb-env-core)
  if "hostedZone" not in stackResources && route53.hostedZoneExists(name=appConfig.domainName):
    // there exists a hostedZone in the user's account that works for this domain name
    // but SaaS Boost didn't create it. so add it in as a parameter: meaning the CloudFormation
    // condition will evaluate to false and no hostedZone will be created (and none will be deleted,
    // since it didn't already exist owned by this template)
    stackParameters.hostedZone = route53.hostedZone(name=appConfig.domainName).id
  else:
    // either hostedZone was created by SaaS Boost or there is no hosted zone in the user account
    // that works for this domain name. in either case we want the "CreateHostedZone" condition
    // to evaluate to true so either:
    //   - the SaaS Boost created hosted zone won't be deleted
    //   or
    //   - a hosted zone will be created by SaaS Boost, since none exist
    // we do this by providing no value for the hostedZone parameter
    stackParameters.hostedZone = "" // the same as having no else in production code 
```

--- 

This is :bug: Bug Report

PoeppingT:
…update.

If you configure a domain name but have no pre-existing HostedZone owned by SaaS Boost to serve that domain name in Route53, SaaS Boost will create one for you. This commit fixes a bug in the OnboardingService when handling AppConfig updates that would delete a HostedZone previously created by this SaaS Boost environment.

Fixes #300 

----

### DIFF
--- a/resources/saas-boost-svc-onboarding.yaml
+++ b/resources/saas-boost-svc-onboarding.yaml
@@ -339,6 +339,7 @@ Resources:
               - cloudformation:CreateChangeSet
               - cloudformation:DeleteStack
               - cloudformation:DescribeStacks
+              - cloudformation:DescribeStackResource
             Resource:
               - !Sub arn:aws:cloudformation:${AWS::Region}:${AWS::AccountId}:stack/sb-${Environment}*
               - !Sub arn:aws:cloudformation:${AWS::Region}:${AWS::AccountId}:transform/saas-boost-app-services-macro
--- a/services/onboarding-service/pom.xml
+++ b/services/onboarding-service/pom.xml
@@ -33,7 +33,7 @@ limitations under the License.
     </licenses>
 
     <properties>
-        <checkstyle.maxAllowedViolations>200</checkstyle.maxAllowedViolations>
+        <checkstyle.maxAllowedViolations>148</checkstyle.maxAllowedViolations>
     </properties>
 
     <build>
--- a/services/onboarding-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/OnboardingService.java
+++ b/services/onboarding-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/OnboardingService.java
@@ -1767,8 +1767,9 @@ public class OnboardingService {
 
         Map<String, Object> appConfig = getAppConfig(context);
         Map<String, Object> services = (Map<String, Object>) appConfig.get("services");
+
         String domainName = (String) appConfig.getOrDefault("domainName", "");
-        String hostedZone = getExistingHostedZone(domainName);
+        String hostedZone = chooseHostedZoneParameter(stackName, domainName, cfn, route53);
 
         // If there's an existing hosted zone, we need to tell the AppConfig about it
         // Otherwise, if there's a domain name, CloudFormation will create a hosted zone
@@ -2185,14 +2186,15 @@ public class OnboardingService {
         return pathPriority;
     }
 
-    protected String getExistingHostedZone(String domainName) {
+    // separate out route53 client for testability
+    protected static String getExistingHostedZone(String domainName, Route53Client route53Client) {
         String existingHostedZone = "";
         if (Utils.isNotEmpty(domainName)) {
             String nextDnsName = null;
             String nextHostedZone = null;
             ListHostedZonesByNameResponse response;
             do {
-                response = route53.listHostedZonesByName(ListHostedZonesByNameRequest.builder()
+                response = route53Client.listHostedZonesByName(ListHostedZonesByNameRequest.builder()
                         .dnsName(nextDnsName)
                         .hostedZoneId(nextHostedZone)
                         .maxItems("100")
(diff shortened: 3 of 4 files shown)


brtrvn (approved):
Confirmed that the SaaS Boost hosted zone for the application domain is preserved during stack update calls.

PR #307, opened 2022-09-27
disable service quotas check in GCR as it is not available

hanrobyn:
- disable service quotas check in GCR as it is not available.


PoeppingT:
Hey @hanrobyn please rework this PR to go against `main` as we're tracking to release China support into the `main` branch rather than just a feature branch.

It looks like you can measure quotas for RDS and ELB using `rds.describeAccountAttributes().accountQuotas();` and `elb.describeAccountLimits();` respectively: could you implement those in your edge case handling?

hanrobyn:
create a new pr https://github.com/awslabs/aws-saas-boost/pull/386 to measure quotas for RDS and ELB, close it

PR #359, opened 2022-10-04
Fix bug attaching multiple CapacityProviders in multiple EC2 services.

### LINKED ISSUE #356: cloudformation stacks for the deployed services failed while onboarding tenants
we've deployed three services on the saas boost app, Client, Proxy and Tellread. But while provisioning a tenant, two of the three cloudformation stacks for the three services came back failed with this error msg and the one for the service tellread succeeded.

sb-dev-ecs-instance-tenant-ccb402c3 already exists in stack arn:aws:cloudformation:us-west-2:042377184132:stack/sb-dev-tenant-ccb402c3-app-tellread-2452RWY4BM53/688f9a80-40ea-11ed-9e81-0a1242b2c085

not really sure how to progress thru this, thanks

saas boost branch: the latest version from master

PoeppingT:
When launching multiple EC2 services the current tenant-onboarding-app CloudFormation stack tries to create a ClusterCapacityAssociations CloudFormation resource for each service. However because attaching CapacityProviders to ECS Clusters is non-atomic only the first ClusterCapacityAssociations will actually succeed.

This commit adds a CustomResource which on CapacityProvider create and delete will attach or detach that CapacityProvider to the existing ECSCluster. Because this attachment is a non-atomic operation (i.e. it involves first getting the existing capacityProviders for the ECS cluster before then overwriting with a new set of capacityProviders) this commit also includes using the existing Onboarding DynamoDB table as a locking mechanism to prevent against illogical overwrites between multiple stacks operating on the same ECS Cluster.

Fixes #356 

----

### DIFF
--- a/layers/cloudformation-utils/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/CloudFormationResponse.java
+++ b/layers/cloudformation-utils/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/CloudFormationResponse.java
@@ -32,6 +32,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 import java.util.function.Function;
 
 public class CloudFormationResponse {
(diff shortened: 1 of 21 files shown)


brtrvn (approved):
Wish there were a cost effective and performant way to persist a lock without overloading the purpose of the onboarding service database. For now, this solution works. Confirmed that multiple EC2 launch type services fail to provision due to capacity provider associations before this fix and works after.

PR #369, opened 2022-10-10
Make version more detailed, move git plugin to parent

PoeppingT:
This PR makes some small changes to how we implement the version string, moving from raw text to a JSON string containing the commit hash, the closest previous tag, and the git description of the repository state. 

This allows us to be more discerning about how we display the version information to the user, and should help the community when diagnosing entered issues by providing more information in the version string.

e.g. the `git.properties` file created on build will look like this
```
  #Generated by Git-Commit-Id-Plugin
  git.closest.tag.name=v2.0.0
  git.commit.id.abbrev=18cc13b
  git.commit.id.describe=v2.0.0-2-g18cc13b-dirty
  git.commit.id.describe-short=v2.0.0-2-dirty
```

and will be translated to JSON like this
```
{
  "tag":"v2.0.0",
  "hash":"18cc13b",
  "describe":"v2.0.0-2-dirty"
}
```

----


brtrvn (commented):
I like it. Any thoughts on using `commit` as the property name instead of `hash` ? Would that make more sense to our community since GitHub tends to refer to "commit" rather than hash.

PoeppingT:
> I like it. Any thoughts on using `commit` as the property name instead of `hash` ? Would that make more sense to our community since GitHub tends to refer to "commit" rather than hash.

I'm fine with doing this. Update incoming

brtrvn (approved):
Looks good

PR #370, opened 2022-10-10
Splitting the system identity provider resources from the admin web app

brtrvn:
Reorganizing the CloudFormation stacks to split the creation of the system identity provider from the creation of the admin web application. This sets us up for much more complicated identity provider provisioning when we start using solutions other than Cognito and also creates more flexibility in how we're doing the admin web app which will solve some customer issues like having to manage a Node toolchain on their workstation and the sequencing of the welcome email and the availability of the admin web app.

----


resources/saas-boost-svc-onboarding.yaml
```diff
@@ -309,6 +309,11 @@ Resources:
               - ecr:DeleteRepository
             Resource:
               - !Sub arn:${AWS::Partition}:ecr:${AWS::Region}:${AWS::AccountId}:repository/*
+          - Effect: Allow
+            Action:
+              - cloudfront:GetDistribution
+            Resource:
+              - !Sub arn:${AWS::Partition}:cloudfront::${AWS::AccountId}:distribution/*
```
PoeppingT, line 316: What do we need this for? I assume this is a preparatory change for the upcoming Keycloak integration PR?

PoeppingT (approved):
One small question, otherwise LGTM

PR #383, opened 2022-10-21
System User Service Updates

brtrvn:
Rename User Service to System User Service to reduce confusion about whether this API is for the SaaS application's tenant users.
Update the service to support multiple implementations based on the system user identity provider.

----


resources/custom-resources/keycloak-setup/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/KeycloakSetup.java
```diff
@@ -156,11 +162,46 @@ protected boolean importRealm(String keycloakHost, String bearerToken, String re
...
+        // Public OAuth app client with PKCE for the admin web app
+        String adminWebAppClientId = Utils.randomString(20);
+        String adminWebAppClientName = realmName + "-admin-webapp-client";
+        LinkedHashMap<String, Object> adminWebAppClient = new LinkedHashMap<>();
+        adminWebAppClient.put("enabled", true);
+        adminWebAppClient.put("protocol", "openid-connect");
+        adminWebAppClient.put("name", adminWebAppClientName);
+        adminWebAppClient.put("clientId", adminWebAppClientId);
+        adminWebAppClient.put("description", "SaaS Boost Admin Web App Client");
+        adminWebAppClient.put("redirectUris", List.of(adminWebAppUrl));
```
PoeppingT, line 174: non-blocking: would be nice to also include `localhost:3000` for testing

resources/saas-boost-idp.yaml
```diff
@@ -253,30 +253,54 @@ Resources:
...
+      Value: !If [UseCognito, !Ref UserPoolClient, !GetAtt keycloak.Outputs.ApiAppClientId]
+  ApiAppClient:
+    Type: AWS::SecretsManager::Secret
     Condition: UseKeycloak
-    Description: URL for the Keycloak install
-    Value: !GetAtt keycloak.Outputs.KeycloakHost
-  KeycloakRealm:
+    Properties:
+      Name: !Sub sb-${Environment}-api-app-client
+      SecretString: !Sub '{"client_name": "${keycloak.Outputs.ApiAppClientName}", "client_id": "${keycloak.Outputs.ApiAppClienId}", "client_secret": "${keycloak.Outputs.ApiAppClientSecret}"'
```
PoeppingT, line 283: sp: `"client_id": "${keycloak.Outputs.ApiAppClien[t]Id}"`

resources/saas-boost-web.yaml
```diff
@@ -129,10 +126,8 @@ Resources:
             Value: !Ref ApiGatewayUrl
           - Name: REACT_APP_CLIENT_ID
             Value: !Ref AdminWebClientId
-          - Name: REACT_APP_COGNITO_USERPOOL
-            Value: !Ref UserPoolId
-          - Name: REACT_APP_COGNITO_USERPOOL_BASE_URI
-            Value: !Ref UserPoolBaseUri
+          - Name: REACT_APP_OIDC_HOST
+            Value: !Ref OidcHost
```
PoeppingT, line 130: this isn't necessary, only the issuer is actually needed. but I can change this when I build on top of this commit.

resources/saas-boost-idp.yaml
```diff
@@ -253,30 +253,54 @@ Resources:
...
+    Value: !If
+      - UseCognito
+      - !Sub saas-boost-${Environment}.auth.${AWS::Region}.amazoncognito.com
+      - !GetAtt keycloak.Outputs.KeycloakHost
+  OidcIssuer:
+    Description: OIDC issuer for System User IdP
+    Value: !If
+      - UseCognito
+      - !Sub saas-boost-${Environment}.auth.${AWS::Region}.amazoncognito.com/${UserPool}
+      - !GetAtt keycloak.Outputs.KeycloakIssuer
```
PoeppingT, line 296: I've only tested with `https://cognito-idp.${AWS::Region}.amazonaws.com/${UserPool}`, have you verified this works the same way?

services/system-user-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/CognitoUserDataAccessLayer.java
```diff
@@ -0,0 +1,291 @@
...
+                }
+            }
+        }
+        return user;
+    }
+
+//    public Map<String, String> getToken(String username, String password) {
+//        Map<String, String> error = new HashMap<>();
+//        Map<String, String> returnData = new HashMap<>();
+//        try {
```
PoeppingT, line 233: I assume this commented out code will be deleted later?

services/system-user-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/CognitoUserDataAccessLayer.java
```diff
@@ -0,0 +1,291 @@
...
+                    .userStatus(response.userStatus())
+                    .userStatus(response.userStatusAsString())
+                    .username(response.username())
+                    .mfaOptions(response.mfaOptions())
+                    .build();
+            user = fromUserType(userType);
+        } catch (SdkServiceException cognitoError) {
+            LOGGER.error("cognito-idp:AdminGetUser", cognitoError);
+            LOGGER.error(Utils.getFullStackTrace(cognitoError));
+            //throw cognitoError;
```
PoeppingT, line 90: why not throw the exception here?

PoeppingT (commented):
General feedback, looks like this is a WIP anyway

resources/custom-resources/keycloak-setup/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/KeycloakSetup.java
PoeppingT: redirectUris should also include a wildcard, e.g. `https://randomstring.cloudfront.net*` to cover redirects towards ui sub-pages, like `https://randomstring.cloudfront.net/dashboard#`

PoeppingT: we also need to configure CORS for access through this client: configuring the [`webOrigins` key](https://www.keycloak.org/docs-api/15.0/rest-api/index.html#_clientrepresentation) for `*` will solve this

services/system-user-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/CognitoUserDataAccessLayer.java
brtrvn: This is old stuff. I'd guess it's because the service calling this DAL is returning a 404 if we can't load a user from the User Pool by username instead of a 5xx. We can dig into what kinds of exceptions are thrown by Cognito and think about what kind of experience that should mean for the client of the API.

PoeppingT (commented):
LGTM, pending implementations for update, enable, and disable in the Keycloak User Service.

installer/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/SaaSBoostInstall.java
```diff
@@ -297,16 +309,245 @@ protected void installSaaSBoost(String existingBucket) {
...
+        String adminWebAppHostedZone = null;
+        String adminWebAppCertificate = null;
+        if (useCustomDomainForAdminWebApp) {
+            if (!Utils.isChinaRegion(AWS_REGION)) {
+                System.out.println("You must provide a verified TLS (SSL) certificate from ACM in the us-east-1 "
+                        + " region to use a custom domain name for the SaaS Boost admin web console.");
+            } else {
+                System.out.println("You must provide a verified TLS (SSL) certificate from the IAM certificate store "
+                        + " to use a custom domain name for the SaaS Boost admin web console.");
+                System.out.println("You must also have an ICP registration for the domain. See https://www.amazonaws.cn/en/about-aws/china/#ICP_ for more information.");
```
PoeppingT, line 424: Looks like this link should be `https://www.amazonaws.cn/en/about-aws/china/#ICP_in_China`

resources/custom-resources/keycloak-setup/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/KeycloakSetup.java
```diff
@@ -126,101 +142,356 @@ public Object handleRequest(Map<String, Object> event, Context context) {
...
+    protected int postUserClientRoleMapping(String keycloakHost, String bearerToken, String realmName, String userId,
+                                            String clientId, Map<String, Object> role) {
+        try {
+            URI endpoint = new URI(keycloakHost + "/admin"
+                    + "/realms/" + realmName
+                    + "/users/" + userId
+                    + "/role-mappings"
+                    + "/clients/" + clientId
+            );
+            String body = Utils.toJson(role);
```
PoeppingT, line 419: This API takes a list of roles. This should be `Utils.toJson(List.of(role));`

PoeppingT (approved):
Two small updates that I will make. We still need to implement disable, enable, and update user, but otherwise LGTM. Good first step.

PR #386, opened 2022-10-24
change service quotas check logic in GCR as QuotaService is not avail…

hanrobyn:
As the QuotaService is not available in GCR, so I measure quotas for RDS and ELB using services itself. 

----


layers/utils/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/Utils.java
```diff
@@ -141,6 +141,17 @@ public static String toJson(Object obj) {
...
+    public static boolean isCnRegion(final String region) {
+        return region.startsWith("cn-");
+    }
+
+    public static String getUrlSuffix(final String region) {
+        if (isCnRegion(region)) {
+            return "amazonaws.com.cn";
+        }
+        return "amazonaws.com";
+    }
```
PoeppingT, line 153: Why not use the `isChinaRegion(String)` and `endpointSuffix(String)` functions already in this file?

services/quotas-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/QuotasServiceDAL.java
```diff
@@ -367,6 +369,77 @@ private Map<String, Double> getQuotas(String serviceCode) {
...
+    private Map<String, Double> getRdsInstancesQuota() {
+        long instances;
+        long clusters;
+        try {
+            clusters = rds.describeAccountAttributes().accountQuotas().stream()
+                    .filter(quota -> quota.accountQuotaName().equals("DBClusters"))
+                    .findFirst().map(AccountQuota::max).orElse(40L);
+            instances = rds.describeAccountAttributes().accountQuotas().stream()
+                    .filter(quota -> quota.accountQuotaName().equals("DBInstances"))
+                    .findFirst().map(AccountQuota::max).orElse(40L);
```
PoeppingT, line 415: Let's cache the result of `rds.describeAccountAttributes().accountQuotas` so we don't waste time doing a second round-trip call to RDS.

services/quotas-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/QuotasServiceDAL.java
```diff
@@ -367,6 +369,77 @@ private Map<String, Double> getQuotas(String serviceCode) {
...
+        return retVals;
+    }
+
+    private Map<String, Double> getELBQuota() {
+        long instances;
+        try {
+            instances = elb.describeAccountLimits().limits().stream()
+                    .filter(x -> x.name().equals("application-load-balancers"))
+                    .findFirst().map(Limit::max).map(Long::valueOf).orElse(50L);
+        } catch (SdkServiceException rdsError) {
```
PoeppingT, line 433: small suggestion: call this `elbError`

PoeppingT (changes requested):

layers/utils/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/Utils.java
hanrobyn: change the code to use isChinaRegion(String) and endpointSuffix(String) functions

services/quotas-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/QuotasServiceDAL.java
hanrobyn: I think if we cache this result, then account changes its quota, we may use the old value.

services/quotas-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/QuotasServiceDAL.java
hanrobyn: I have renamed it.

services/quotas-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/QuotasServiceDAL.java
hanrobyn: I use a accountQuotas field to cache the result.

PoeppingT (approved):
LGTM, thanks for the submission!

PR #387, opened 2022-10-24
change for CloudFront do not support managed policies in GCR

hanrobyn:
Use deprecated ForwardedValues property in GCR as CloudFront managed cache policy and managed origin request policy are not currently supported in GCR.

----

### DIFF
--- a/resources/saas-boost.yaml
+++ b/resources/saas-boost.yaml
@@ -83,6 +83,9 @@ Conditions:
   ShouldCreateMacroResources: !Equals [!Ref CreateMacroResources, 'true']
   UseCognito: !Equals [!Ref SystemIdentityProvider, 'COGNITO']
   UseKeycloak: !Equals [!Ref SystemIdentityProvider, 'KEYCLOAK']
+  IsCNRegion: !Or
+    - !Equals [ !Sub '${AWS::Region}', 'cn-north-1' ]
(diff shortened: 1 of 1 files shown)


resources/saas-boost.yaml
```diff
@@ -83,6 +83,9 @@ Conditions:
   ShouldCreateMacroResources: !Equals [!Ref CreateMacroResources, 'true']
   UseCognito: !Equals [!Ref SystemIdentityProvider, 'COGNITO']
   UseKeycloak: !Equals [!Ref SystemIdentityProvider, 'KEYCLOAK']
+  IsCNRegion: !Or
+    - !Equals [ !Sub '${AWS::Region}', 'cn-north-1' ]
+    - !Equals [ !Sub '${AWS::Region}', 'cn-northwest-1' ]
```
PoeppingT, line 88: It is more forward-compatible to instead check that `${AWS::Partition}` is equal to `aws-cn`. Adding more regions is more likely than adding new partitions, and we don't want this to break if new regions are added.

PoeppingT (changes requested):

resources/saas-boost.yaml
hanrobyn: changed it to `IsCNRegion: !Equals [!Ref AWS::Partition, 'aws-cn']`

PoeppingT (approved):
LGTM, thanks for the submission!

PR #395, opened 2022-10-31
Rds bootstrap batching

brtrvn:
Batch the execution of the SQL bootstrap file when it contains many statements to try and avoid memory issues and transaction commit size performance problems.

----


resources/custom-resources/rds-bootstrap/src/test/java/com/amazon/aws/partners/saasfactory/saasboost/RdsBootstrapTest.java
```diff
@@ -42,4 +42,32 @@ public void testSqlScanner() {
         }
         assertEquals(6, sql.size());
     }
+
+    @Test
+    public void testBatch() {
+        InputStream bootstrapSQL = Thread.currentThread().getContextClassLoader().getResourceAsStream("large.sql");
+        Scanner sqlScanner = new Scanner(bootstrapSQL, "UTF-8");
+        sqlScanner.useDelimiter(Pattern.compile(RdsBootstrap.SQL_STATEMENT_DELIMITER));
```
PoeppingT, line 50: This logic appears to just test a copy-paste of the logic in RdsBootstrap. A more complete way to test this would be to refactor this specific functionality into an alternate function that takes a `Connection`, a `Statement`, and an `InputStream` as input and makes the requisite calls against the `Connection` and `Statement` reading from the `InputStream`. The test can mock the calls made to those classes and ensure that `addBatch` is only called a maximum of `MAX_SQL_BATCH_SIZE` times before `commit` is called, and that each batch is genuinely only that large.

PoeppingT (approved):
Small comment re: testing, but we can address that in an upcoming PR

PR #419, opened 2022-11-16
Modify for china region can not download keycloak-metrics-spi from gi…

hanrobyn:
Can not download keycloak-metrics-spi.jar from github in china region, so I put it in the project.

----


PoeppingT:
Hi @hanrobyn , thanks for the PR! This jar adds metrics Prometheus-scrapeable endpoints to Keycloak and is a jar we included following best practices in Keycloak. At this time the metrics are not actually used, so this PR can be changed to remove the jar entirely.

First, can you provide more detail on why the jar cannot be downloaded from GitHub in China region? What restrictions are there on downloads? It appears that you are able to download jars from Maven, since you were able to build the project. If this same binary were hosted at a different location, would the download succeed?

hanrobyn:
> 

The connection from China region to Github is slow and unstable, So download the whole jar always cause a timeout error, but the access to Maven is ok.  
But I can not find this jar in Maven.

hanrobyn:
close it for clear branch. 
and create a new pr for it: https://github.com/awslabs/aws-saas-boost/pull/422

PR #422, opened 2022-11-17
Remove keycloak-metrics-spi.jar when build keycloak

hanrobyn:
----
Remove the unused jar when build keycloak, as we can't download it from github in China region.

As the connection from China region to Github is slow and unstable, so download the whole jar always cause a timeout error, but the access to Maven is ok.
But I can not find this jar in Maven.


PoeppingT (approved):
Looks good, thanks for the contribution!

PR #427, opened 2022-11-17
Switch from SSE-KMS to SSE-SQS for SQS encryption at rest

brtrvn:
Replace the KMS key with SQS owned encryption keys for encryption at rest of the SQS queues.

This change allows EventBridge to directly interact with SQS instead of requiring a customer managed KMS key with a special key policy.

----


PoeppingT (approved):
LGTM

PR #428, opened 2022-11-18
Implement update, enable, and disable for Keycloak System User Service.

PoeppingT:
This commit contains a number of changes for the Keycloak implementation of the System User Service, as well as small fixes for the User view in the admin website.

----


resources/custom-resources/keycloak-setup/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/KeycloakSetup.java
```diff
@@ -270,7 +285,72 @@ protected int postRealm(String keycloakHost, String bearerToken, Map<String, Obj
...
+                LOGGER.error("Expected HTTP_CREATED ({}) from group create, but got {}",
+                        HttpURLConnection.HTTP_CREATED, response.statusCode());
+                throw new RuntimeException("Unexpected error while creating group: " + groupName);
+            }
+        } catch (URISyntaxException | IOException | InterruptedException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    protected Map<String, Object> findGroupByName(String keycloakHost, String bearerToken,
```
brtrvn, line 319: Might be able to do GET /{realm}/groups?search={groupName} here instead of pulling back all the groups and iterating over them

services/system-user-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/keycloak/KeycloakUserDataAccessLayer.java
```diff
@@ -0,0 +1,157 @@
...
+    // VisibleForTesting
+    static UserRepresentation updateUserRepresentation(SystemUser user, UserRepresentation keycloakUser) {
+        if (user != null) {
+            keycloakUser.setUsername(user.getUsername());
+            keycloakUser.setFirstName(user.getFirstName());
+            keycloakUser.setLastName(user.getLastName());
+            keycloakUser.setEnabled(user.getActive());
+            keycloakUser.setEmail(user.getEmail());
+            keycloakUser.setEmailVerified(Boolean.TRUE.equals(user.getEmailVerified()));
+            keycloakUser.setCreatedTimestamp(LocalDateTime.now().toInstant(ZoneOffset.UTC).toEpochMilli());
```
brtrvn, line 149: Now that we have update implemented, should we wrap this in an if/else to see if our incoming SystemUser object already has a created timestamp -- so we don't turn created into modified.
```java
if (user.getCreated() == null) {
    keycloakUser.setCreatedTimestamp(LocalDateTime.now().toInstant(ZoneOffset.UTC).toEpochMilli());
} else {
    keycloakUser.setCreatedTimestamp(user.getCreated().toInstant(ZoneOffset.UTC).toEpochMilli());
}
```

services/system-user-service/src/test/java/com/amazon/aws/partners/saasfactory/saasboost/keycloak/KeycloakTestUtils.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import static org.junit.Assert.assertEquals;
+
+public final class KeycloakTestUtils {
+    public static void assertUserListsEqual(List<UserRepresentation> expected, List<UserRepresentation> actual) {
+        // if expected and actual are not equal in existence
+        if (expected == null ^ actual == null) {
```
brtrvn, line 32: Wow XOR for the win!
Replace with `((expected == null && actual == null) || (expected != null && actual != null))` for clarity?

brtrvn (commented):
Any sense in incorporating resilience4j retry in these 3rd party calls to REST APIs? Like we did in CloudFormationUtils?
```java
RetryConfig exponentialBackoffWithJitter = RetryConfig.custom()
            .maxAttempts(MAX_RESPONSE_ATTEMPTS)
            .retryExceptions(CloudFormationResponseException.class)
            .intervalFunction(
                    // 1s, 2s, 4s, 8s, 16s retries with random jitter
                    IntervalFunction.ofExponentialRandomBackoff(1000L, 2.0d, 0.6d)
            )
            .build();
```
This helps set us up for moving the "API" stuff into a layer that we could share between the System User Service and the custom resource that sets up the Keycloak install via CloudFormation. We should be thinking about how this pattern will be repeated with other identity providers like Auth0 and Ping Identity.

resources/custom-resources/keycloak-setup/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/KeycloakSetup.java
PoeppingT: Nice callout, I'll do that instead.

services/system-user-service/src/test/java/com/amazon/aws/partners/saasfactory/saasboost/keycloak/KeycloakTestUtils.java
PoeppingT: > Any sense in incorporating resilience4j retry in these 3rd party calls to REST APIs? Like we did in CloudFormationUtils?

I like the idea in concept, but what would we retry? The only thing I can think of would be `429 Too Many Requests`, since we have no indication that [any other HTTP code](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_client_errors) would necessarily be retryable. In addition I can find no documentation or code reference that indicates a throttling mechanism in Keycloak.

PoeppingT:
@brtrvn 

> Any sense in incorporating resilience4j retry in these 3rd party calls to REST APIs? Like we did in CloudFormationUtils?

I like the idea in concept, but what would we retry? The only thing I can think of would be `429 Too Many Requests`, since we have no indication that [any other HTTP code](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_client_errors) would necessarily be retryable. In addition I can find no documentation or code reference that indicates a throttling mechanism in Keycloak.

brtrvn (approved):
Looks good. Thanks for the updates!

Issue #446, opened 2022-12-05
Bill Event Process gives error


raman231:
Bill Event Process lambda giving following error

class java.lang.Integer cannot be cast to class java.lang.Long (java.lang.Integer and java.lang.Long are in module java.base of loader 'bootstrap'): java.lang.ClassCastException
java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.Long (java.lang.Integer and java.lang.Long are in module java.base of loader 'bootstrap')
	at com.amazon.aws.partners.saasfactory.metering.billing.ProcessBillingEvent.handleRequest(ProcessBillingEvent.java:118)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.base/java.lang.reflect.Method.invoke(Unknown Source)

### Reproduction Steps

Invoke "Artificial Meter Event for Billing" from the provided sample java application.

### What did you expect to happen?

Billing event should process successfully

### What actually happened?

Billing event process giving error in logs.

### Environment

  - **AWS Region  :us-east-1
  - **AWS SaaS Boost Version :v2.0.0-48
  - **Workload OS  (Linux or Windows) :Linux

### Other

--- 

This is :bug: Bug Report

PR #447, opened 2022-12-05
Fix for Issue #446

raman231:
----


PoeppingT (changes requested):
Thanks for the contribution @raman231 ! In this case, you're effectively changing the model for the `BillingEvent` in integration for user applications, which would constitute a backwards-incompatible change for existing users integrated with the BillingEvent. In this case why not change [the sample app](https://github.com/awslabs/aws-saas-boost/blob/main/samples/java/src/main/java/com/amazon/aws/partners/saasfactory/controller/MeteringController.java) to send a Long value instead of an Integer?

```
        int meterVal = Integer.valueOf(count); // could be a Long
        long startTimeMillis = System.currentTimeMillis();

        try {
            ObjectNode systemApiRequest = MAPPER.createObjectNode();
            systemApiRequest.put(TENANT_ID, tenantId);
            systemApiRequest.put(PRODUCT_CODE, productCode);
            systemApiRequest.put(QUANTITY, meterVal);
```

raman231:
> Thanks for the contribution @raman231 ! In this case, you're effectively changing the model for the `BillingEvent` in integration for user applications, which would constitute a backwards-incompatible change for existing users integrated with the BillingEvent. In this case why not change [the sample app](https://github.com/awslabs/aws-saas-boost/blob/main/samples/java/src/main/java/com/amazon/aws/partners/saasfactory/controller/MeteringController.java) to send a Long value instead of an Integer?
> 
> ```
>         int meterVal = Integer.valueOf(count); // could be a Long
>         long startTimeMillis = System.currentTimeMillis();
> 
>         try {
>             ObjectNode systemApiRequest = MAPPER.createObjectNode();
>             systemApiRequest.put(TENANT_ID, tenantId);
>             systemApiRequest.put(PRODUCT_CODE, productCode);
>             systemApiRequest.put(QUANTITY, meterVal);
> ```

Changes made as suggested. Thanks

PR #448, opened 2022-12-06
Manually signout in Cognito environments with a /logout redirect.

### LINKED ISSUE #441: Signout from admin web app does not work with Cognito
Because Cognito does not follow the common log out pattern of OIDC, signing out of the admin web app does not actually end the session. It will redirect you to the sign in UI, but if your session tokens are not yet expired, you will be automatically signed back in as the same user.

### Reproduction Steps

Choose Cognito as your system user identity provider during install
Sign into SaaS Boost and update your temporary password
Sign out of SaaS Boost
Click the sign in button
Instead of being prompted for a username and password, you are redirected back to the home page as the last signed in user

After signing out, being prompted for username and password at next sign in

### Environment

  - **AWS Region  :** any that supports Cognito

### Other

--- 

This is :bug: Bug Report

PoeppingT:
Since Cognito doesn't support the `end_session_endpoint` OIDC parameter relied upon by the oidc-client-ts library we use in the admin application, we need special logic for Cognito environments when implementing logout to make a manual call to the /logout endpoint.

Fixes #441 

----

### DIFF
--- a/client/web/src/App.js
+++ b/client/web/src/App.js
@@ -31,6 +31,7 @@ App.propTypes = {
 
 function App(props) {
   const oidcAuth = props.oidcAuth
+  const signout = props.signout
   if (props.authState === 'signedIn') {
     return (
       <Provider store={store}>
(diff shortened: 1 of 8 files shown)


brtrvn (approved):
LGTM

PR #450, opened 2022-12-13
SaaS Boost API access via OAuth client credentials grant

brtrvn:
When we upgraded the System User Service to use OAuth `authorization_code` grant for access to the SaaS Boost API instead of Amazon Cognito native authorization, we removed the lesser-known private API `/token` endpoint that returned an `Identity Token` for the SaaS Boost admin user.

This PR generates an application client with a secret to be used for `client_credentials` grant against the SaaS Boost API. It will return an `Access Token` with the equivalent permissions as the SaaS Boost admin user. This token can be used as a bearer token for any of the public SaaS Boost API resources.

The client's details are stored in Secrets Manager at `/saas-boost/${ENVIRONMENT}/API_APP_CLIENT`. Any client code wanting to perform computer-to-computer invocation of the SaaS Boost API will need to be given explicit allow permissions via IAM to access that secret.

**_DO NOT_** use this secret in a public client such as a JavaScript web app or a mobile application. This should only be used in backend server side code that is authorized.

----


resources/saas-boost.yaml
```diff
@@ -846,6 +846,7 @@ Resources:
         CustomDomainName: !If [HasSystemIdPCustomDomain, !Ref SystemIdentityProviderDomain, '']
         CustomDomainHostedZone: !If [HasSystemIdPCustomDomain, !Ref SystemIdentityProviderHostedZone, '']
         CustomDomainCertificate: !If [HasSystemIdPCustomDomain, !Ref SystemIdentityProviderCertificate, '']
+        ApiGatewayUrl: !Sub https://${core.Outputs.SaaSBoostPublicApi}.execute-api.${AWS::Region}.${AWS::URLSuffix}/${PublicApiStage} #!GetAtt publicapi.Outputs.PublicApiGatewayEndpoint
```
PoeppingT, line 849: Why can't we just use `!GetAtt` here?

brtrvn: Because it would make a circular dependency with the saas-boost-public-api.yaml stack

PoeppingT (approved):
Thanks, LGTM! Excited to see what we can do with automated testing using this change.

PR #451, opened 2022-12-15
Tenant update race condition

### LINKED ISSUE #243: tenant is successfully deployed but not hostname is shown in the tenant page
Tenant is successfully deployed, tenantService log is found(updating the hostname to ddb), but no items are found in ddb and no hostname is shown in the tenant page.

ELB was there, listener was set correctly, ecs service was running as well.. only hostname field was not shown in ddb..
  
  
  
  
Here is what I found in the loggroup that seems like ddb put event was successfully done
  

### Reproduction Steps

Idk how I can reproduce this error, but during the test, I encountered this symptom twice..

hostname should be saved into ddb so that pooled front-end application can fetch the elb endpoint from ddb.

### Environment

  - **AWS Region  :**

### Other

--- 

This is :bug: Bug Report

brtrvn:
Use DynamoDB `updateItem` instead of `getItem` -> `updateItem` when processing tenant events to avoid unexpected loss of results when concurrent invocations step on each other.

This PR also adds IAM permissions for tag management of ECR repos to resolve an unreported error during initial app config save when the ECR repository is generated for each service. 

Fixes #243 

----


PoeppingT (approved):
Manually tested, LGTM.

Issue #452, opened 2022-12-30
OnboardingService Error


raman231:
Getting following error while onboarding a tenant - 

ERROR OnboardingService - cloudformation::createStack failed
software.amazon.awssdk.services.cloudformation.model.CloudFormationException: Unable to fetch parameters [/aws/service/ami-windows-latest/Windows_Server-20H2-English-Core-ECS_Optimized/image_id] from parameter store for this account. (Service: CloudFormation, Status Code: 400

### Reproduction Steps

Onboard tenant through admin ui

### What did you expect to happen?

Tenant should be onboarded successfully

### What actually happened?

Tenant onboarding failed

### Environment

  - **AWS Region  :us-east-1
  - **AWS SaaS Boost Version :Version v2.1.0-2-dirty@89a5af4
  - **Workload OS  (Linux or Windows) :Linux

### Other

--- 

This is :bug: Bug Report

PR #458, opened 2023-01-04
Fix for Issue #452

raman231:
Windows 10, version 20H2 has reached end of servicing.  Refer - https://learn.microsoft.com/en-us/windows/release-health/status-windows-10-20h2

And there's no reference to /aws/service/ami-windows-latest/Windows_Server-20H2-English-Core-ECS_Optimized/image_id now. Removing this image since its no longer usable.

----


PoeppingT (approved):
While this PR does not address removing 20H2 from the AppConfig options or the UI, this will unblock users with an AppConfig that does not use that version.

PR #460, opened 2023-01-05
Rds options orderable instance fix

brtrvn:
Work around the unexpected case of a database engine version returned from [describe-db-engine-versions](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/rds/describe-db-engine-versions.html) not returning from [describe-orderable-db-instance-options](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/rds/describe-orderable-db-instance-options.html).

This is likely due to "eventual consistency" in RDS as older versions are deprecated, but it was an uncaught exception for our custom resource and was breaking install.

----


PoeppingT (approved):
Manually tested, LGTM.

PR #463, opened 2023-01-06
Support S3 storage for tenants as AppConfig extension.

PoeppingT:
This commit adds support for configuring S3 support as part of your SaaS Boost application. If configured a single S3 bucket will be created for all your tenants to read/write from, only allowable under a specific prefix passed as an environment variable to the ECS Task.

Related environment variables for use:
- S3_BUCKET: the name of the s3 bucket created
- S3_ROOT_PREFIX: the prefix under which this tenant is allowed to read/write files

----


client/web/src/settings/components/ObjectStoreSubform.js
```diff
@@ -0,0 +1,59 @@
...
+        <Row className="mt-3">
+          <Col xs={12}>
+            <Card>
+              <CardHeader>Object Storage</CardHeader>
+              <CardBody>
+                <SaasBoostCheckbox
+                  id={this.props.formikServicePrefix + '.provisionObjectStorage'}
+                  name={this.props.formikServicePrefix + '.provisionObjectStorage'}
+                  disabled={this.props.isLocked}
+                  label="Provision an S3 bucket for the application."
```
brtrvn, line 44: Could consider exposing Bucket Versioning as a provisioning choice

resources/custom-resources/app-services-macro/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/ApplicationServicesMacro.java
```diff
@@ -52,8 +55,146 @@ public Map<String, Object> handleRequest(Map<String, Object> event, Context cont
...
+
+    protected static String updateTemplateForS3(final Map<String, Object> templateParameters,
+            Map<String, Object> template) {
+        if (templateParameters.containsKey("Environment")
+                && templateParameters.containsKey("LoggingBucket")) {
+            String environmentName = (String) templateParameters.get("Environment");
+            // Bucket Resource
+            Map<String, Object> s3Resource = s3Resource(environmentName,
+                    (String) templateParameters.get("LoggingBucket"));
+            ((Map<String, Object>) template.get("Resources")).put("S3Extension", s3Resource);
```
brtrvn, line 123: Prefer we call this Resource `storage` or `objects` (lowercase) instead of `S3Extension` to try and match a little better the pattern we use for other provisioned S3 buckets and to make its use maybe a little more clear.

resources/custom-resources/clear-s3-bucket/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/ClearS3Bucket.java
```diff
@@ -46,6 +46,8 @@ public Object handleRequest(Map<String, Object> event, Context context) {
         final String requestType = (String) event.get("RequestType");
         Map<String, Object> resourceProperties = (Map<String, Object>) event.get("ResourceProperties");
         final String bucket = (String) resourceProperties.get("Bucket");
+        final String passedPrefix = (String) resourceProperties.get("Prefix");
+        final String prefix = passedPrefix == null ? "" : passedPrefix;
```
brtrvn, line 50: Consider overly paranoid `final String prefix = passedPrefix == null ? "" : passedPrefix..replaceFirst("^/", "");`

resources/tenant-onboarding-app.yaml
```diff
@@ -229,6 +229,10 @@ Parameters:
   OnboardingDdbTable:
     Description: Internal DDB table used to store Onboarding Metadata
     Type: String
+  S3ExtensionBucket:
```
brtrvn, line 232: Other naming options that kinda match existing patterns better
`ObjectStorageBucket` `StorageBucket` `ObjectsBucket`

resources/tenant-onboarding-app.yaml
```diff
@@ -482,40 +492,30 @@ Resources:
...
+              - Name: DB_PORT
+                Value: !If [ProvisionRDS, !Ref RDSPort, !Ref AWS::NoValue]
+              - Name: DB_NAME
+                Value: !If [ProvisionRDS, !Ref RDSDatabase, !Ref AWS::NoValue]
+              - Name: DB_USER
+                Value: !If [ProvisionRDS, !Ref RDSUsername, !Ref AWS::NoValue]
+              - Name: S3_BUCKET
+                Value: !If [UseS3, !Ref S3ExtensionBucket, !Ref AWS::NoValue]
+              - Name: S3_ROOT_PREFIX
+                Value: !If [UseS3, !Sub '${ServiceResourceName}/${TenantId}/', !Ref AWS::NoValue]
```
brtrvn, line 518: I don't think we should include the tenant id as part of the prefix ENV variable. It doesn't match the existing S3 patterns we have and if we get to a place where we have pooled compute resources, the SaaS application will have to set the tenant id at runtime.

services/onboarding-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/OnboardingService.java
```diff
@@ -1029,6 +1029,10 @@ protected void handleOnboardingBaseProvisioned(Map<String, Object> event, Contex
                         templateParameters.add(Parameter.builder().parameterKey("RDSPort").parameterValue(dbPort.toString()).build());
                         templateParameters.add(Parameter.builder().parameterKey("RDSDatabase").parameterValue(dbDatabase).build());
                         templateParameters.add(Parameter.builder().parameterKey("RDSBootstrap").parameterValue(dbBootstrap).build());
+                        templateParameters.add(Parameter.builder()
+                                .parameterKey("S3ExtensionBucket")
+                                .parameterValue(getSetting(context, "EXTENSION_BUCKET"))
```
brtrvn, line 1034: If we change the same to storage or objects or something else, we should be consistent throughout and change it in param store and wherever else

resources/tenant-onboarding-app.yaml
```diff
@@ -431,6 +436,11 @@ Resources:
                   - s3:GetObjectVersion
                 Resource:
                   - !Sub arn:${AWS::Partition}:s3:::{{resolve:ssm:/saas-boost/${Environment}/RESOURCES_BUCKET}}/tenants/${TenantId}/*
+              - Effect: Allow
+                Action:
+                  - s3:*
+                Resource:
+                  - !Sub arn:${AWS::Partition}:s3:::${S3ExtensionBucket}/${ServiceResourceName}/${TenantId}/*
```
brtrvn, line 443: This is too permissive. There are bucket level actions we don't want the SaaS application to be able to modify. SaaS Boost owns the provisioning and configuration of this bucket.

```yaml
- Effect: Allow
  Action:
    - s3:ListBucket
    - s3:ListBucketVersions
  Resource:
    - !Sub arn:${AWS::Partition}:s3:::${S3ExtensionBucket}
  Condition:
    StringLike:
      s3:prefix:
        - !Sub ${ServiceResourceName}/${TenantId}/*
- Effect: Allow
  Action:
    - s3:GetBucket*
    - s3:GetAccelerateConfiguration
    - s3:GetAnalyticsConfiguration
    - s3:GetEncryptionConfiguration
    - s3:GetIntelligentTieringConfiguration
    - s3:GetInventoryConfiguration
    - s3:GetLifecycleConfiguration
    - s3:GetMetricsConfiguration
  Resource:
    - !Sub arn:${AWS::Partition}:s3:::${S3ExtensionBucket}
- Effect: Allow
  Action:
    - s3:GetObject*
    - s3:PutObject*
    - s3:DeleteObject*
    - s3:AbortMultipartUpload
    - s3:ListMultipartUploadParts
  Resource:
    - !Sub arn:${AWS::Partition}:s3:::${S3ExtensionBucket}/${ServiceResourceName}/${TenantId}/*
```

brtrvn (changes requested):
Overall looking pretty good.

We can consider adding Bucket Versioning as an optional configuration choice (it would apply to all services though in our current implementation).

I think we should remove reference to "extension" in some of the naming and maybe even "s3" in some of the naming (is there any other object store?) and stick to "storage" or "objects" generically.

Note that as currently implemented, if you try to list the bucket contents when it's "empty" you'll get a `403` access denied because the prefix you're allowed to list doesn't exist until you put at least 1 object. This may surprise users (surprised me). I think the only way to resolve this would be with a Custom Resource that on CREATE calls

```java
s3.putObject(request -> request
    .bucket(bucket)
    .key(prefix)
    , RequestBody.empty()
);
```

Essentially creating an empty "folder" like the AWS console does so that calls to list the bucket don't fail unexpectedly.

Also, we would want to double check our clear s3 bucket logic to make sure it deleted the prefix as well as everything underneath it.

client/web/src/settings/components/ObjectStoreSubform.js
PoeppingT: As discussed, we'll leave this for posterity and address as necessary for customers.

brtrvn (approved):
Thanks for the changes. Looks good. I still might prefer App Storage over Tenant Storage, but it's not important enough to change right now I don't think.

Simplistic example of how this could be used:

```java
String bucket = System.getenv("S3_BUCKET");
String bucketPrefix = System.getenv("S3_ROOT_PREFIX");
String tenantId = System.getenv("TENANT_ID");

// Create the prefix for this tenant
String prefix = bucketPrefix + tenantId + "/";
LOGGER.info("Calling list-objects-v2 on s3://{}/{}", bucket, prefix);

try {
    S3Client s3 = S3Client.builder().build();
    ListObjectsV2Response response = s3.listObjectsV2(request -> request
        .bucket(bucket)
        .prefix(prefix)
    );
    List<String> objects = response.contents()
        .stream()
        .map(obj -> obj.key())
        .collect(Collectors.toList()
    );
    return objects;
} catch (S3Exception s3Error) {
    LOGGER.error("S3 error", s3Error);
}
```

PR #470, opened 2023-01-30
fix(sec): upgrade com.fasterxml.jackson.core:jackson-databind to 2.14.0-rc1

1derian:
### What happened？
There are 1 security vulnerabilities found in com.fasterxml.jackson.core:jackson-databind 2.13.4.1
- [CVE-2022-42003](https://www.oscs1024.com/hd/CVE-2022-42003)

### What did I do？
Upgrade com.fasterxml.jackson.core:jackson-databind from 2.13.4.1 to 2.14.0-rc1 for vulnerability fix

### What did you expect to happen？
Ideally, no insecure libs should be used.

### The specification of the pull request
[PR Specification](https://www.oscs1024.com/docs/pr-specification/) from OSCS

### DIFF
--- a/functions/authorizer/pom.xml
+++ b/functions/authorizer/pom.xml
@@ -80,7 +80,7 @@ limitations under the License.
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
-            <version>2.13.4.1</version>
+            <version>2.14.0-rc1</version>
         </dependency>
         <dependency>
             <groupId>com.auth0</groupId>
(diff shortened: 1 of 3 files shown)


brtrvn (commented):
Looks like this issue does not impact us (the effected setting is off by default and we have not enabled it). Also, I'm hesitant to jump to a release candidate over a long term supported release of Jackson since it's such a critical shared library. Also note that the fix is a breaking fix so we'd need to test accordingly.

https://github.com/FasterXML/jackson-databind/issues/3590


We will keep our eye on this issue and respond when the dust has settled.

PoeppingT:
According to the CVE record on https://nvd.nist.gov/vuln/detail/CVE-2022-42003 this issue is only present in Jackson before 2.13.4.1, but is fixed in 2.13.4.1. 

Closing this issue as a false alarm.

PR #476, opened 2023-02-07
Fix bug where deleted service validation errors stayed at top of AppConfig page.

PoeppingT:
Previously, if you create a service then delete it immediately after creating (or have a service with validation errors that you delete) the validation warning at the top of the page would stay there. This did not affect submission, since clicking `Submit` would re-run the form validation discounting the deleted service, but was a bad customer experience. This commit forcibly removes both the deleted service from the Formik errors object and Formik values object, which allows us to remove the unnecessary and complicated 'tombstone' concept.

----


PoeppingT:
Closing this in favor of a new PR to fold in changes not pointed to by the current commit message.

Issue #474, opened 2023-02-03
services names swapped 


goodbyedavid:
### What actually happened?

After creating two services, I defined two tiers for each service. 
Once the tiers were created, it is noticed service names were swapped.

### Environment

  - **AWS Region  : cn-northeast-1
  - **AWS SaaS Boost Version : v2-lastest
  - **Workload OS  (Linux or Windows) : Linux

### Other
<img width="586" alt="截屏2023-02-03 09 45 47" src="https://user-images.githubusercontent.com/3408755/216491623-eab853cf-8de4-4496-b842-6d5f0cb7ee1b.png">

<img width="576" alt="截屏2023-02-03 09 46 16" src="https://user-images.githubusercontent.com/3408755/216491703-3e0d35a8-5c8f-4b42-9f89-320383bd76a7.png">

--- 

This is :bug: Bug Report

PoeppingT:
Hi @goodbyedavid , thanks for the bug report! To address this issue we'll need a little more information:
1. In your picture you show that `main` has the Service Name field as `feature1`. Does that `main` service have the configurations of `main` or `feature1`?
2. What is the exact order of operations you took on your SaaS Boost environment? How much of the application config did you create before creating tiers? How did the application config change once you created the tiers?
3. Behind the hood, this application configuration is stored as a JSON blob in AWS SSM Parameter Store. Go to your AWS Console, navigate to Systems Manager, click on Parameter Store on the left, and find the parameters named `/saas-boost/${environment}/app/${main/feature1}/SERVICE_JSON`. Do those JSON parameters match what you expect?

I suspect that this is simply a UI bug, and that the application configuration under the hood is still as you expect: but answers to these questions will help us get to the bottom of the issue.

PoeppingT:
#477 may fix this issue, but until we have reproduction we cannot know for sure. @goodbyedavid have you been able to reproduce this issue?

PR #477, opened 2023-02-08
Improve Application Config UX

PoeppingT:
This commit improves on multiple issues with the existing application configuration page in the admin UI:
1. deleting a service now actually deletes the errors for that service from the top banner
2. creating a new service automatically opens the accordion banner for that new service
3. the validation error banner won't be shown for new services until any error is made (assuming there are not existing errors in other services)
4. the accordion name is now controlled directly by the form, rather than an alternate state
5. removes the unnecessarily complicated `tombstone` concept by manually deleting the service from the form

4/ in the above list should also address #474, but until we have reproduction we cannot know for sure.

----


brtrvn (approved):
LGTM

PR #478, opened 2023-02-09
Fix line graphs in metrics dashboards

PoeppingT:
The period data returned by our metrics service includes Z as a timezone, which indicates UTC. Adding the extra UTC made the data unparseable, leading to NaN:NaN being represented as our periods.

Before fix. Note `NaN-NaN NaN:NaN` as the time period on the left.
> Before fix

After fix. Note we actually have data.
> After fix

----


brtrvn (approved):
LGTM

PR #480, opened 2023-02-10
Update scripts must not fail if Lambda function does not already exist.

PoeppingT:
When updating from older versions of SaaS Boost to newer ones, we rely on each Lambda function's update.sh script to build the code and update any relevant Lambda functions in the AWS account. If any of these update.sh scripts fail, an entire SaaS Boost update will fail, since it assumes something has gone wrong with building or uploading that code.

However, in the case we add a new Lambda function, the existing Lambda function may not already exist until the CloudFormation stack is updated, which is something the Installer executes after running each update.sh script. This change makes these update scripts not fail in that case.

----

### DIFF
--- a/functions/authorizer/update.sh
+++ b/functions/authorizer/update.sh
@@ -48,8 +48,7 @@ fi
 
 aws s3 cp target/$LAMBDA_CODE s3://$SAAS_BOOST_BUCKET/$LAMBDA_STAGE_FOLDER/
 
-FUNCTIONS=("sb-${ENVIRONMENT}-authorizer" 
-        )
+eval FUNCTIONS=\$\("aws --region $MY_AWS_REGION lambda list-functions --query 'Functions[?starts_with(FunctionName, \`sb-${ENVIRONMENT}-authorizer\`)] | [].FunctionName' --output text"\)
 
 for FUNCTION in ${FUNCTIONS[@]}; do
 	#echo $FUNCTION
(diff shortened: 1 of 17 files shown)


brtrvn (commented):
Since `--query` is a client side action, and `list-functions` is a paginated call, we either need to include `--no-paginate` or use a different mechanism like `get-function` and then check for an error code of `254` (or just looking at `STDERR`)

In the case of layers, functions, and custom resources, we're only ever looking for a single function, so maybe it makes more sense (and may be more performant) to just use `get-function` to check for existence.

For the services' update scripts, we may have to add `--no-paginate`

PoeppingT:
> For the services' update scripts, we may have to add `--no-paginate`

To test this I created 1000 lambda functions, and it looks like we don't need to do anything special with pagination for services.

With 1000 functions labelled `example{0...1000}`
```
aws lambda list-functions --query 'Functions[?ends_with(FunctionName, `"0"`)] | [].FunctionName' --output text | sed 's/\t/\n/g' | wc -l
     101
```
and
```
aws lambda list-functions --query 'Functions[?starts_with(FunctionName, `example`)] | [].FunctionName' --output text | sed 's/\t/\n/g' | wc -l
    1001
```

Not entirely sure why the normal service update scripts work without the `sed` command, but I'm willing to take that at face value. I agree that it might make sense to switch to `get-function` for performance reasons, but for reference scanning 1000 functions only takes ~2 seconds, which is far less than the amount of time it takes to build the code and update each function anyway, so I would argue it isn't worth the extra effort of doing error checking for the minimal performance gain.

brtrvn (commented):
It works without replacing `\t` with `\n` because we're assigning the value to a Bash array -- which will take whitespace separated strings for each item in the array. You can get the length of the array with `${#FUNCTIONS[@]}` if you want to print it out or test for `0`.

brtrvn (approved):

PR #485, opened 2023-02-14
Allow user to specify desired HostedZone in AppConfig UI.

### LINKED ISSUE #257: DNS hosted zone should be registered when we set app config
DNS hosted zone is registered during tenant onboarding process. Even if I already have same domain name registered in route53 as hosted zone, saas boost register that domain name again in route 53 so it ended up having two same domain registered in route53.

I think even because of ssl cert verification, hosted zone should be registered when we set app config, so that the admin can do DNS verification before having tenant onboarding process. 

If hosted zone is registered in onboarding process, then NS(name server) should be changed in domain purchase site and dns verification should be done after first tenant is onboard.. which seems to be not right order..

### Environment

  - **AWS Region  :**

### Other

--- 

This is :bug: Bug Report

### LINKED ISSUE #461: Tenant access error
Thanks for your update.  

SaasBoost 2.1.0-8 installed
I have successfully created a tenant.
But Created  tenant access 404 error
Is this my problem?

S3-Backet-accesslog?
Saasboost admin colsole-access log is recorded

PoeppingT:
This commit removes the capability for SaaS Boost to automatically create a Route53 HostedZone for your Application Domain Name. Instead, users should have their DNS configuration set up before configuring their SaaS Boost application with a domain name.

Additionally, users can now reconfigure their Application's domain name if they have no tenants onboarded.

Fixes #257.
Fixes #461.

----


resources/saas-boost.yaml
```diff
@@ -907,7 +903,6 @@ Resources:
         PrivateApiStage: !Ref PrivateApiStage
         ApplicationServices: !Ref ApplicationServices
         DomainName: !Ref ApplicationDomainName
```
brtrvn, line 905: The DomainName parameter isn't used anywhere and we can remove it

services/onboarding-service/src/main/java/com/amazon/aws/partners/saasfactory/saasboost/OnboardingService.java
```diff
@@ -1835,19 +1835,9 @@ protected void handleUpdateInfrastructure(Map<String, Object> event, Context con
         Map<String, Object> services = (Map<String, Object>) appConfig.get("services");
 
         String domainName = (String) appConfig.getOrDefault("domainName", "");
```
brtrvn, line 1837: We can remove both domain name and hosted zone from this method. They aren't part of the CloudFormation stack anymore.

brtrvn (changes requested):

brtrvn (approved):
Looks good!

PR #486, opened 2023-02-23
Make FSx Active Directory Configuration Optional

brtrvn:
Using FSx for NetApp ONTAP on Linux hosts does not require the Linux host to be joined to an Active Directory domain like it does for Windows hosts. This PR makes the Active Directory configuration of the FSx storage virtual machine optional. You no longer have to provision a managed Active Directory instance to use FSx ONTAP with Linux.

----


PoeppingT (approved):
LGTM

PR #487, opened 2023-02-24
Move filesystem configuration to services, performance characteristics to tier.

PoeppingT:
This commit refactors the way that AppConfig is saved for filesystems: moving the filesystem configuration object to be service specific and allowing only the filesystem performance characteristics to be configurable by tier, replacing the existing logic where all filesystem configurations were configurable by tier.

This commit also reworks the UI to support the above changes as well as changing the Onboarding service to correctly select from the AppConfig object.

----


brtrvn (approved):
Looks great!

PR #491, opened 2023-03-01
Fix Metric Line Graphs, fix Sysuser create redirect.

PoeppingT:
This commit fixes two small bugs:
1. The previous fix for fixing the metrics line graphs was incomplete. This commit fixes it for good by using the Date calculation if it can be successfully converted into a Date object, and just using the returned Date otherwise.
2. When we moved to calling things "system-user-service", the redirect from the UserCreateContainer was missed, resulting in a broken breadcrumb when creating a user. Changing the redirect location fixes this issue.

----


brtrvn (commented):
LGTM - notice that the `Compute` graph _for all tenants_ shows individual markers on the line and `Requests` doesn't. Both appear to show individual markers when you drill into a single tenant view.

brtrvn (approved):

PR #493, opened 2023-03-03
Support enabling ECS Exec for individual services at creation time.

PoeppingT:
This commit adds support for the ECS Exec feature of ECS. Enabling ECS Exec in the Admin UI for a given service will create the ECS Service with the ExecuteCommand option and allowed permissions for ExecuteCommand, which will let admins run commands on tenant service containers for debugging purposes. Those permissions are also explicitly denied for services where ECS Exec is not enabled.

----

### DIFF
--- a/client/web/src/settings/ApplicationComponent.js
+++ b/client/web/src/settings/ApplicationComponent.js
@@ -161,6 +161,7 @@ export function ApplicationComponent(props) {
       name: thisService?.name || serviceName,
       path: thisService?.path || '/*',
       public: thisService ? thisService.public : true,
+      ecsExecEnabled: thisService ? thisService.ecsExecEnabled : false,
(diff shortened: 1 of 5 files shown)


brtrvn (commented):
I don't think it will, but we should make sure that the updated IAM policy doesn't break SSM Session Manager connections (connect through the AWS Console without a key pair) to the underlying EC2 host.

PoeppingT:
> I don't think it will, but we should make sure that the updated IAM policy doesn't break SSM Session Manager connections (connect through the AWS Console without a key pair) to the underlying EC2 host.

Confirmed that with ECS Exec enabled I can connect to a running EC2 instance using the Session Manager connect option.

brtrvn (approved):

PR #495, opened 2023-03-07
Fix overwritten database bootstrap file in AppConfig UI.

### LINKED ISSUE #245: An issue on update config page
Given the condition that there is no onboarded tenant yet, after setting an app config and submitting that including bootstrap file, when the user tries to update the db version, then the post request post 'null' value of bootstrap file ever since 

my file should be there..

### What actually happened?
null value is there..

### Environment

  - **AWS Region  :**

### Other

--- 

This is :bug: Bug Report

PoeppingT:
The database bootstrap files are uploaded to S3 using presigned URLs for the service resources bucket generated by the Settings service. The UI will first update AppConfig then upload any files to their presigned locations when the new AppConfig is returned from Settings service containing the presigned URLs. The Settings service will then receive an event when a new object is put in that resources bucket and will save the bootstrap filename into the AppConfig.

This commit makes it so that the AppConfig UI page mirrors the bootstrapFilename back to the Settings service if no new bootstrap file was selected in the file upload component. This means that the same bootstrap filename will be saved in Settings service backend, no new presigned url will be generated, and the configuration will retain the indication that the same bootstrap file should be used for subsequent database onboardings.

Fixes #245.

----

### DIFF
--- a/client/web/src/settings/ApplicationContainer.js
+++ b/client/web/src/settings/ApplicationContainer.js
@@ -163,6 +163,7 @@ export function ApplicationContainer(props) {
       }
       // clean tiers if present
       if (!!filesystem?.tiers) {
+        let cleanedTiers = {}
         for (let tierName in filesystem.tiers) {
           let {
             weeklyMaintenanceDay,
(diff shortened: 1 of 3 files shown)


brtrvn (approved):
LGTM

PR #498, opened 2023-03-15
Refactor appConfig compute and support setting min/max for EC2 AutoScaling Group.

PoeppingT:
This commit contains multiple changes:
1. refactors the ServiceConfig object to contain a top-level "compute" structure, defined with type, that contains all definitions related to a specific type of compute. Right now this includes container information, since future types of compute (e.g. Lambda) may not necessarily be defined by containers. Note that this does not contain path or public/private, since those describe the application running on the compute, not the compute itself. This commit also includes changes to the UI, Onboarding Service, auxiliary functionality (i.e. workload-deploy and core-stack-listener functions), and the sample application build scripts to handle this change with respect specifically to the container repository information.
2. Supports configuring the minimum and maximum EC2 AutoScaling Group size when the EC2 launch type is configured. This will either be when the operating system is Linux and that launch type is selected, or if the operating system is Windows. This commit also includes changes in the Settings and Onboarding services, as well as tenant-onboarding stacks and the UI to support easy configuration of this new setting.
3. This commit also includes a small bugfix in the UI: when creating a new service on the AppConfig page some changed logic to support FileSystem Tier configurations was leading to a failure.

----

### DIFF
--- a/client/web/src/dashboard/DashboardComponent.js
+++ b/client/web/src/dashboard/DashboardComponent.js
@@ -234,16 +234,16 @@ export const DashboardComponent = (props) => {
                             <dl>
                               <dt>ECR Repository</dt>
                               <dd>
-                                {isEmpty(service.containerRepo)
+                                {isEmpty(service?.compute?.containerRepo)
                                   ? 'Creating...'
-                                  : service.containerRepo}{' '}
+                                  : service?.compute?.containerRepo}{' '}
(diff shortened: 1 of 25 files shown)


brtrvn (approved):
This is a great change for logical consistency in the app config and sets us up to be able to more easily model other types of compute in the future such as EKS and Lambda.

PR #500, opened 2023-03-23
Create a managed AD per tenant, not per environment.

PoeppingT:
----


resources/tenant-onboarding-managed-ad.yaml
```diff
@@ -33,56 +33,60 @@ Parameters:
...
+        ExcludePunctuation: true
   ADDirectory:
     Type: AWS::DirectoryService::MicrosoftAD
     Properties:
       Name: !Sub ${Environment}.${AWS::Region}.saasboost.com
       Edition: !Ref Edition
-      Password: !Sub '{{resolve:ssm-secure:${ADPasswordParam}}}'
-      ShortName: !Sub sb-${Environment}
+      Password: !Sub '{{resolve:secretsmanager:${ADPassword}::::}}'
+      ShortName:
```
brtrvn, line 53: Short Name can't exceed 15 characters (NETBIOS limit)
WIth `sb-` and `-tenantId` that only leaves 3 characters for the environment. Suggest changing it to just Tenant-`tenantId` which should always fit.

brtrvn (changes requested):

brtrvn (changes requested):
We need to fix the UserData scripts in tenant-onboarding-app.yaml to fetch the AD credentials from the proper place now that we've moved them. The UserData script is currently failing for Windows hosts so a container expecting a mounted file system will not stabilize.

brtrvn (approved):

PR #513, opened 2023-05-05
Fix onboarding new tenants with EFS

uhinze:
Tenant onboarding with EFS is currently broken.

My research so far:
- https://docs.aws.amazon.com/efs/latest/ug/API_CreateFileSystem.html is being called including a list of tags. I can see this call in CloudTrail
- Under the hood, this call possibly used to do a https://docs.aws.amazon.com/efs/latest/ug/API_CreateTags.html which is deprecated, and is now doing a https://docs.aws.amazon.com/efs/latest/ug/API_TagResource.html. For this, the onboarding role lacks permission. 

I noticed this issue when running through this workshop: https://catalog.workshops.aws/saasboost/en-US/fast-lab/lab1/part5

Error in Cloudformation is: 
`
User: arn:aws:sts::972741443277:assumed-role/sb-boost-onboarding-svc-role-us-west-2/sb-boost-onboarding-events is not authorized to perform: elasticfilesystem:TagResource on the specified resource (Service: Efs, Status Code: 403, Request ID: c4120ca5-7782-4248-8134-4c750eaf98dd)” (RequestToken: cb6d5312-b8d3-f5f6-b898-826bb8c282b5, HandlerErrorCode: GeneralServiceException)
`

----


brtrvn:
Thanks for the diagnosis and fix. This must be an intermittent issue with the CloudFormation fleet. They are probably rolling out updates in phases. I can successfully provision tenants with _either_ the `elasticfilesystem:CreateTags` permission _or_ the `elasticfilesystem:TagResource` in us-west-2.

brtrvn (changes requested):
Since I can reproduce this failure, but I can see the proposed change work, I suggest for now that we _add_ the `TagResource` permission to the policy and keep the `CreateTags` permission until we can confirm changes on the EFS/CloudFormation side.

brtrvn (approved):
LGTM. Thank you for the contribution!

PR #546, opened 2026-07-08
Add NoEcho to sensitive CloudFormation password parameters

andrewpowell:
## Summary

Adds the `NoEcho` property to the two CloudFormation input parameters that are consumed as a database `Password` but did not mask their value. This follows AWS CloudFormation best practice for sensitive input parameters (masking the value in the console, stack events, and `DescribeStacks` API output).

| Template | Parameter |
|---|---|
| `resources/tenant-onboarding-rds.yaml` | `RDSPasswordParam` |
| `resources/saas-boost-metrics-analytics.yaml` | `MetricUserPasswordSSMParameter` |

`NoEcho: true` only masks the value in the console/API/logs; `!Ref` resolution is unaffected, so this is non-breaking.

## Verification

Validated with **cfn-lint 1.46**:
- Before: two `W2501` findings ("Parameter ... used as Password, therefore NoEcho should be True").
- After: **zero `W2501`** findings across all templates in `resources/`.

The credential-*retrieval* side already follows best practice throughout the repo — secrets are stored in Secrets Manager / SSM SecureString and read via dynamic references (`{{resolve:secretsmanager:...}}` / `{{resolve:ssm-secure:...}}`).

## Notes / out of scope

- `resources/saas-boost-keycloak.yaml` exposes an OAuth2 client secret via a stack **Output** (`ApiAppClientSecret`). CloudFormation Outputs cannot take `NoEcho`; addressing it would require a Lambda + nested-stack-contract change and is intentionally left out of this focused PR.

### DIFF
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -6,10 +6,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout PR
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
 
       - name: Cache Maven packages
-        uses: actions/cache@v2
+        uses: actions/cache@v4
         with:
           path: ~/.m2
           key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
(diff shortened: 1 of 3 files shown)


andrewpowell:
@brtrvn would you be able to review this? Thanks!

andrewpowell:
@tobuck-aws tagging you as well for review — thanks!

ranjith-aws:
LGTM

brtrvn:
Note that these are not the password, they are the name of the SSM parameter... So, no credentials are leaking anywhere. We could probably change the name of the template parameter to avoid cfn-lint thinking it's a password... Something like CredsParam...

